PR #103, opened 2022-08-26
Hostname template

aakoshh:
This PR allows the `testnet` command that set up `-v` number of nodes to use a hostname template in `--starting-ip-address` rather than an IP address. The template only works for `{i}` and replaces it with `0, 1, 2, ...`. 

This is so that we can use predictable hostnames when deploying to AWS, when we don't know the IP addresses up front. 

The IP/hostname becomes part of the `memo` field of the transaction that creates the validator and is signed, so it can't be modified later. The `genutil` module of the Cosmos SDK [collects](https://github.com/cosmos/cosmos-sdk/blob/3e492d4bcae76b715bed89215c70fad13aa9f20a/x/genutil/collect.go#L38) these [memos](https://github.com/cosmos/cosmos-sdk/blob/3e492d4bcae76b715bed89215c70fad13aa9f20a/x/genutil/collect.go#L129) and puts them in the `persistent_peers` of `.testnets/node0/babylond/config/config.toml`, and for `node1` etc.

Although, for provisioning a network, it should also be possible to just edit the `config.toml` files and modify the seed and/or persistent node list, surely they don't have to match the transaction memos. They are probably there as a convenience mechanism conjured up by the Cosmos developers.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+.testnet
+.testnets
--- a/Makefile
+++ b/Makefile
@@ -435,7 +435,7 @@ localnet-build-dlv:
 
 localnet-build-nodes:
 	$(DOCKER) run --rm -v $(CURDIR)/.testnets:/data babylonchain/babylond \
-			  testnet init-files --v 4 -o /data --starting-ip-address babylonnode{i} --keyring-backend=test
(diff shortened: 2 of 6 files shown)


vitsalis:
Closing due to inactivity.

PR #300, opened 2023-02-02
zoneconcierge API: pagtinating chain IDs API

SebastianElvis:
Fixes [BM-485](https://babylon-chain.atlassian.net/browse/BM-485)

This PR adds pagination for chain IDs API in order to rule out a potential DoS vector. 

Context: https://github.com/babylonchain/babylon-api/pull/41#discussion_r1094333071


proto/babylon/zoneconcierge/query.proto
```diff
@@ -76,11 +76,16 @@ message QueryHeaderResponse {
 }
 
 // QueryChainListRequest is request type for the Query/ChainList RPC method
-message QueryChainListRequest {}
+message QueryChainListRequest {
+  // pagination defines whether to have the pagination in the response
```
gitferry, line 80:
```suggestion
  // pagination defines whether to have the pagination in the request
```

proto/babylon/zoneconcierge/query.proto
```diff
@@ -76,11 +76,16 @@ message QueryHeaderResponse {
...
+  cosmos.base.query.v1beta1.PageRequest pagination = 1;
+}
 
 // QueryChainListResponse is response type for the Query/ChainList RPC method
 message QueryChainListResponse {
   repeated string chain_ids = 1;
+  // pagination defines the pagination in the response
```
gitferry, line 87: Do we need to add a docstring to say that the ids are returned in the ascending alphabet order? Maybe we can add that in several places.

x/zoneconcierge/keeper/grpc_query_test.go
```diff
@@ -32,33 +32,38 @@ func FuzzChainList(f *testing.F) {
...
 			} else {
 				chainID = datagen.GenRandomHexStr(30)
-				expectedChainIDs = append(expectedChainIDs, chainID)
+				allChainIDs = append(allChainIDs, chainID)
 			}
 			header := datagen.GenRandomIBCTMHeader(chainID, 0)
 			hooks.AfterHeaderWithValidCommit(ctx, datagen.GenRandomByteArray(32), header, false)
 		}
 
+		limit := datagen.RandomInt(len(allChainIDs)-1) + 1
```
gitferry, line 50: Why do we want to minus 1 here? Do we want the limit to be always at least 1 less than the number of chain IDs?

x/zoneconcierge/keeper/grpc_query_test.go
```diff
@@ -32,33 +32,38 @@ func FuzzChainList(f *testing.F) {
...
-		sort.Strings(expectedChainIDs)
+		// sort them and assert actualChainIDs is a subset of allChainIDs
+		// NOTE: chain IDs in KVStore are in different orders than allChainIDs
+		sort.Strings(allChainIDs)
 		sort.Strings(actualChainIDs)
-		require.Equal(t, len(expectedChainIDs), len(actualChainIDs))
-		for i := 0; i < len(expectedChainIDs); i++ {
-			require.Equal(t, expectedChainIDs[i], actualChainIDs[i])
-		}
+		require.Equal(t, limit, uint64(len(actualChainIDs)))
```
gitferry, line 65: I feel that this only covers the case in the first page. Maybe it would be more comprehensive to cover more pages. Please see [this](zoneconcierge-paginate-chainlist).

gitferry (approved):
Awesome! One bit is that maybe we want to test more pages when testing pagination

x/zoneconcierge/keeper/grpc_query_test.go
SebastianElvis: Good catch, I found `-1` is not necessary here. I wanted to make the limit to be at least 1 (since limit cannot be 0) and at most `len(allChainIDs)`, but it seems that `RandomInt(x)` returns at most `x-1`. Removed `-1` then.

x/zoneconcierge/keeper/grpc_query_test.go
SebastianElvis: Hmm maybe ensuring the pagination API takes effect is enough here? The pagination functionality is already well-tested by Cosmos SDK. As long as it takes effect in basic cases, then it means our code uses pagination API correctly, so probably fine?

gitferry: Agreed. Probably fine

PR #306, opened 2023-02-15
prepare-genesis cmd: Add flags related to inflation

vitsalis:
These flags will help us control the inflation rate when we create a genesis file for testnet purposes.


SebastianElvis (approved):
Thanks! Do we want to control `GoalBonded` in mint params as well? My feeling is that Babylon's potential tokenomics may require a larger `GoalBonded` than other Cosmos zones since most tokens in Babylon will be used for validation rather than on-chain applications.

vitsalis:
Good idea, added this along with blocks per year.

PR #307, opened 2023-02-16
feat: Checkpointing/Add fee estimation in sending BLS-sig tx

### LINKED ISSUE #168: Fees as a parameter for the BlsSig transaction
See the TODO [here](https://github.com/babylonchain/babylon/blob/b24cade923719ed757ab7b906b3788664d2b9124/x/checkpointing/keeper/bls_signer.go#L43). We also need to identify a proper value to set as a default. Maybe we could also investigate using the "auto" value that automatically identifies the proper gas to spend. cc @gitferry

gitferry:
This PR fixes #168 

The fee of sending a BLS-sig tx was hardcoded. This PR estimates the fee by calculating the gas. This PR also adds two gas-related settings, `gas-price` and `gas-adjustment` in the `app.toml` to calculate the actual `fee=calculated_gas * gas_adjustment * gas_price`.

This PR also removed the test of `SendBlsSig` as the gas calculation is not implemented for testing. Instead, this PR is well tested using [babylon-deployment](https://github.com/babylonchain/babylon-deployment).

### DIFF
--- a/client/tx/tx.go
+++ b/client/tx/tx.go
@@ -3,12 +3,9 @@ package tx
 import (
 	"fmt"
 
-	appparams "github.com/babylonchain/babylon/app/params"
 	"github.com/cosmos/cosmos-sdk/client"
-	"github.com/cosmos/cosmos-sdk/client/flags"
 	sdktx "github.com/cosmos/cosmos-sdk/client/tx"
 	sdk "github.com/cosmos/cosmos-sdk/types"
-	"github.com/spf13/pflag"
 )
 
 func SendMsgToTendermint(clientCtx client.Context, msg sdk.Msg) (*sdk.TxResponse, error) {
@@ -22,16 +19,18 @@ func SendMsgsToTendermint(clientCtx client.Context, msgs []sdk.Msg) (*sdk.TxResp
 		}
 	}
 
-	// TODO make the fee dynamic
-	fs := pflag.NewFlagSet("", pflag.ContinueOnError)
-	fs.String(flags.FlagFees, "", "Fees to pay along with transaction; eg: 10ubbn")
-	fee := fmt.Sprintf("100%s", appparams.BaseCoinUnit)
-	err := fs.Set(flags.FlagFees, fee)
-	if err != nil {
-		return nil, err
-	}
-
-	txf := sdktx.NewFactoryCLI(clientCtx, fs)
+	// TODO gasAdjustment and gasPrices could be moved to the checkpoinging params
+	txf := sdktx.Factory{}.
+		WithTxConfig(clientCtx.TxConfig).
+		WithAccountRetriever(clientCtx.AccountRetriever).
+		WithKeybase(clientCtx.Keyring).
+		WithChainID(clientCtx.ChainID).
+		WithFeeGranter(clientCtx.FeeGranter).
+		WithFeePayer(clientCtx.FeePayer).
+		WithGasAdjustment(1.5).
+		WithGasPrices("0.01ubbn")
+	// WithFees("100ubbn")
+	// sdktx.NewFactoryCLI()
 
 	return BroadcastTx(clientCtx, txf, msgs...)
 }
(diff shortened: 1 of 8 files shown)


types/signer_config.go
```diff
@@ -17,3 +21,66 @@ func ParseKeyNameFromConfig(opts servertypes.AppOptions) string {
...
+	valueInterface := opts.Get("signer-config.gas-price")
+	if valueInterface == nil {
+		return "", errors.New("signer gas price should be provided in options")
+	}
+	gasPrice, err := cast.ToStringE(valueInterface)
+	if err != nil {
+		return "", errors.New("signer gas price should be valid string")
+	}
+
+	_, err = sdk.ParseCoinNormalized(gasPrice)
```
vitsalis, line 35: Shouldn't it be more than `0`?

vitsalis, line 35: Should we check that this is more than 0?

cmd/babylond/cmd/custom_babylon_config.go
```diff
@@ -15,16 +19,22 @@ type BtcConfig struct {
...
+		CheckpointTag: txformat.DefaultMainTagStr,
 	}
 }
 
 func defaultSignerConfig() SignerConfig {
-	return SignerConfig{KeyName: ""}
+	return SignerConfig{
+		KeyName:       "",
+		GasPrice:      fmt.Sprintf("0.01%s", params.BaseCoinUnit),
+		GasAdjustment: 1.2,
```
vitsalis, line 30: Should we make those constants?

vitsalis (commented):
Much cleaner, thanks! The build is failing though

vitsalis (approved):
Nice! Out of curiosity, how much fees do the BLS transactions need now?

client/tx/tx.go
```diff
@@ -22,16 +22,16 @@ func SendMsgsToTendermint(clientCtx client.Context, msgs []sdk.Msg) (*sdk.TxResp
...
-	fs := pflag.NewFlagSet("", pflag.ContinueOnError)
-	fs.String(flags.FlagFees, "", "Fees to pay along with transaction; eg: 10ubbn")
-	fee := fmt.Sprintf("100%s", appparams.BaseCoinUnit)
-	err := fs.Set(flags.FlagFees, fee)
-	if err != nil {
-		return nil, err
-	}
-
-	txf := sdktx.NewFactoryCLI(clientCtx, fs)
+	gasPrice, gasAdjustment := types.MustGetGasSettings(clientCtx.HomeDir, clientCtx.Viper)
```
SebastianElvis, line 25: A small optimisation could be passing `gasPrice` and `gasAdjustment` via arguments of `SendMsgToTendermint` rather than loading them upon each `SendMsgToTendermint` invocation. The BLS signer, which has access to these two arguments in memory, is the only component that uses this function at the moment.

SebastianElvis (approved):
Nicely done! This is an important feature 💯

client/tx/tx.go
gitferry: Thanks for the suggestion. Maybe I miss something but I'm not sure what's the benefit of passing those two arguments via `SendMsgToTendermint` because `gasPrice` and `gasAdjustment` depend on `clientCtx`, which is already an argument to `SendMsgToTendermint`. Wouldn't that be redundant if we also pass `gasPrice` and `gasAdjustment`?

SebastianElvis: Here `MustGetGasSettings` needs to read the config file to get `gasPrice` and `gasAdjustment`, so everytime we trigger `SendMsgToTendermint` it will read the file system. I was suggesting to read it once when initialising BLS signer and reuse it throughout the execution. Just a small optimisation, your call

gitferry: That would be an excellent optimization, thanks! Implementing this optimization would cause a bit of breaking changes as the gas settings should be added as part of the BLS signer. Recorded in #311 for now.

gitferry:
@vitsalis My observation during a local deployment was around 6k. So if we set the `gasPrice=0.01ubbn`, that would be around `60ubbn` for one BLS-sig.

Issue #174, opened 2022-10-24
Accept encoding config in `InitPrivSigner` function instead of creating it.


KonradStaniec:
Currently InitPrivSigner function - https://github.com/babylonchain/babylon/blob/main/app/utils.go#L62 creates its own encoding config. This should be rather accepted as argument as there should be only one config per app.

gitferry:
Resolved by #308

PR #308, opened 2023-02-17
chore: Keep a single encoding config in the application

gitferry:
This PR addresses #174 to keep a single encoding config in the application other than creating a new one in `InitPrivSigner`.


app/test_helpers.go
```diff
@@ -242,7 +244,8 @@ func SetupPrivSigner() (*PrivSigner, error) {
 	if err != nil {
 		return nil, err
 	}
-	privSigner, _ := InitPrivSigner(client.Context{}, ".", kr)
+	encodingCfg := appparams.MakeTestEncodingConfig()
```
vitsalis, line 247: Isn't this function deprecated? See [here](https://github.com/babylonchain/babylon/blob/dev/app/encoding.go#L11)

SebastianElvis: Since the encoding config never changes except when adding a new module, perhaps we need to create a unified function that returns the encoding config, or even make it a singleton. I wrote such a function [in rpc-client](https://github.com/babylonchain/rpc-client/blob/dev/config/codec.go#L26-L35). Perhaps this should instead be done in Babylon repo?

gitferry: @vitsalis It is indeed deprecated but it seems that it is the only way to get the global encoding config. It should be only called once during the lifecycle of the validator and this PR is to make sure of that.

@SebastianElvis I think `appparams.MakeTestEncodingConfig()` sets global `ModuleBasics` and returns the encoding config. So, maybe we can just call `appparams.MakeTestEncodingConfig()` in `rpc-client`?

SebastianElvis: > So, maybe we can just call appparams.MakeTestEncodingConfig() in rpc-client?

The rpc-client introduces this new function exactly because `MakeTestEncodingConfig` is deprecated 🤣 so how about we formalise `appparams.MakeTestEncodingConfig()` a bit to make something like `appparams.MakeEncodingConfig()`?

KonradStaniec: from what I see in Juno and Osmosis, each have similar setup to ours. They just do do not call their function `MakeTestEncodingConifg` but `MakeEncodingConfig` and those functions are not deprecated. In general, I wonder why this function (`MakeTestEncodingConfig` in proto.go) is deprecated ? from file history it seems it was marked as such from the start.

My general sentiment here would be to check what others are doing (osmosis, juno, gaia?) and do the same, maybe we could change the naming also i.e drop the `Test` from name and remove deprecation mark if we do not know why it is there.

From what i understand the only real requirement here is that `MakeEncodingConfig` should be called only once per application initialisation.

gitferry:
@vitsalis @SebastianElvis @KonradStaniec Thanks you guys for your input. I followed Osmosis' approach to rename `MakeTestEncodingConfig` into `MakeEncodingConfig` and added a getter to globally get the singleton of the `encodingConfig`. This PR ensures that the `MakeEncodingConfig` is only called once during the lifecycle of the application

app/app_test.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 func TestBabylonBlockedAddrs(t *testing.T) {
-	encCfg := MakeTestEncodingConfig()
+	encCfg := MakeEncodingConfig()
```
SebastianElvis, line 14: Here should we use `GetEncodingConfig` or `MakeEncodingConfig`? Looks like the former one is usable since the package has already finished initialising `encodingConfig` when reaching this line.

app/app_test.go
```diff
@@ -48,7 +48,7 @@ func TestGetMaccPerms(t *testing.T) {
 }
 
 func TestUpgradeStateOnGenesis(t *testing.T) {
-	encCfg := MakeTestEncodingConfig()
+	encCfg := MakeEncodingConfig()
```
SebastianElvis, line 51: ditto

app/test_helpers.go
```diff
@@ -77,7 +80,7 @@ type SetupOptions struct {
 
 func setup(withGenesis bool, invCheckPeriod uint) (*BabylonApp, GenesisState) {
 	db := dbm.NewMemDB()
-	encCdc := MakeTestEncodingConfig()
+	encCdc := MakeEncodingConfig()
```
SebastianElvis, line 83: same here, looks like `GetEncodingConfig` is usable?

SebastianElvis (commented):
Thanks for the fix! Looks like many `MakeEncodingConfig` invocations can be replaced by `GetEncodingConfig`. Maybe I miss something. Otherwise lgtm!

app/app_test.go
gitferry: Thanks for your comment. I just realized that we should only use `GetEncodingConfig` because the global variable `encodingConfig` is initiated when the `app` package is loaded. So I made `madeEncodingConfig` private and only use `GetEncodingConfig` externally.

SebastianElvis (approved):
Thanks, now we have a single source of encoding config 💯

KonradStaniec (approved):

PR #309, opened 2023-02-17
fix gas cost of insert header

KonradStaniec:
Currently most costly babylon msg is Insert Header msg from btc light client. The difference is in order of magnitude range. The main reason is that btclightclient exposes hooks for header events to which btccheckpoint subscribes. With each header update it check status of every checkpoint of every non finalized epoch. This mean that gas cost of this operation is around O(numberOfNonFinlizedEpochs * numberOfCheckpointsInEachNonFinalizedEpoch) (+ some some cost of light client itself).

This pr proposes to move all checking logic into EndBlock callback and to use local transient store to check if btc light client have been updated during current block execution.

This:
- reduces gas cost of inserting btc header to only the gas costs of btc light client itself (operations on transient store are super cheap)
- reduces gas cost of transactions which have multiple Insert Header messages i.e if one inserts 50 btc headers at once, ultimately what matters is the state at the end of Babylon block.

Gas cost of inserting headers in e2e tests have been reduced almost two times.

### DIFF
--- a/x/btccheckpoint/keeper/keeper.go
+++ b/x/btccheckpoint/keeper/keeper.go
@@ -20,6 +20,7 @@ type (
 	Keeper struct {
 		cdc                   codec.BinaryCodec
 		storeKey              storetypes.StoreKey
+		tstoreKey             storetypes.StoreKey
 		memKey                storetypes.StoreKey
 		paramstore            paramtypes.Subspace
 		btcLightClientKeeper  types.BTCLightClientKeeper
@@ -60,6 +61,7 @@ const (
 func NewKeeper(
 	cdc codec.BinaryCodec,
 	storeKey,
+	tstoreKey,
 	memKey storetypes.StoreKey,
 	ps paramtypes.Subspace,
 	bk types.BTCLightClientKeeper,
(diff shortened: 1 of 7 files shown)


x/btccheckpoint/keeper/keeper.go
```diff
@@ -347,6 +350,21 @@ func (k Keeper) OnTipChange(ctx sdk.Context) {
 	k.checkCheckpoints(ctx)
 }
 
+func (k Keeper) setBtcLightClientUpdated(ctx sdk.Context) {
+	store := ctx.TransientStore(k.tstoreKey)
+	store.Set(types.GetBtcLigtClientUpdatedKey(), []byte{1})
```
vitsalis, line 355:
```suggestion
	store.Set(types.GetBtcLightClientUpdatedKey(), []byte{1})
```

x/btccheckpoint/types/keys.go
```diff
@@ -18,16 +18,21 @@ const (
...
 	QuerierRoute = ModuleName
 
+	TStoreKey = "transient_btccheckpoint"
+
 	// MemStoreKey defines the in-memory store key
 	MemStoreKey = "mem_btccheckpoint"
 
 	LatestFinalizedEpochKey = "latestFinalizedEpoch"
+
+	btcLighClientUpdated = "btcLightClientUpdated"
```
vitsalis, line 28:
```suggestion
	btcLightClientUpdated = "btcLightClientUpdated"
```

vitsalis (approved):
That's a very clean fix, thanks!

SebastianElvis (approved):
This is a pretty good use case for transient store!

PR #313, opened 2023-02-21
chore: Add fuzz test of addBlsSig

### LINKED ISSUE #279: Add test for adding a BLS-sig transaction
The test should cover the following phases in one:
* A chain is bootstrapped with a set of genesis validators
* After some epochs, a validator is created and bonded
* Verify that this validator submits a BLS sig at the end of the next epoch

gitferry:
This PR fixed #279.

It added comprehensive fuzz tests of the checkpointing module handling `MsgAddBlsSig`, covering below cases:
1. a BLS signature is successfully accumulated and the checkpoint remains ACCUMULATING
2. a BLS signature is successfully accumulated and the checkpoint is changed to SEALED
3. a BLS signature is rejected if the checkpoint is not ACCUMULATING 
4. a BLS signature is rejected if it signs a wrong last_commit_hash
5. a BLS signature is rejected if the signer is not in the checkpoint's validator set
6. a BLS signature is rejected if the signature is invalid
7. a BLS signature is rejected if the checkpoint does not exist


x/checkpointing/keeper/utils.go
```diff
@@ -0,0 +1,7 @@
+package keeper
+
+import sdk "github.com/cosmos/cosmos-sdk/types"
+
+func GetSignBytes(epoch uint64, hash []byte) []byte {
```
SebastianElvis, line 5: Since this function is not a part of the keeper, maybe it can be moved to `checkpointing/types`?

SebastianElvis (commented):
Thanks, testing BLS signer is definitely not easy! The CI seems to fail though

SebastianElvis (approved):

x/checkpointing/keeper/utils.go
gitferry: Great idea, thanks! Done

PR #318, opened 2023-02-23
chore: Add balance check before inserting MsgWrappedCreateValidator into the epoch message queue

### LINKED ISSUE #317: Lack of balance check before inserting MsgWrappedCreateValidator into the message queue
This issue is found in our Discord discussion. Due to the lack of balance check, a `MsgWrappedCreateValidator` might fail when execution at the end of an epoch. We should add a check before inserting it into the message queue.

gitferry:
This PR fixed #317 and added relevant fuzz tests. To support the check of delegator balance, this PR added the `BankKeeper` to the epoching module

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -10,7 +10,6 @@ import (
 
 	nodeservice "github.com/cosmos/cosmos-sdk/client/grpc/node"
 
-	bbn "github.com/babylonchain/babylon/types"
 	"github.com/gorilla/mux"
 	"github.com/rakyll/statik/fs"
 	"github.com/spf13/cast"
@@ -19,7 +18,8 @@ import (
 	tmos "github.com/tendermint/tendermint/libs/os"
 	dbm "github.com/tendermint/tm-db"
 
-	appparams "github.com/babylonchain/babylon/app/params"
+	bbn "github.com/babylonchain/babylon/types"
+
 	"github.com/cosmos/cosmos-sdk/baseapp"
 	"github.com/cosmos/cosmos-sdk/client"
 	"github.com/cosmos/cosmos-sdk/client/grpc/tmservice"
@@ -87,8 +87,15 @@ import (
 	upgradekeeper "github.com/cosmos/cosmos-sdk/x/upgrade/keeper"
 	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
 
+	appparams "github.com/babylonchain/babylon/app/params"
+
 	// unnamed import of statik for swagger UI support
 	_ "github.com/cosmos/cosmos-sdk/client/docs/statik"
+
+	storetypes "github.com/cosmos/cosmos-sdk/store/types"
+	govclient "github.com/cosmos/cosmos-sdk/x/gov/client"
+	govv1 "github.com/cosmos/cosmos-sdk/x/gov/types/v1"
+	govv1beta1 "github.com/cosmos/cosmos-sdk/x/gov/types/v1beta1"
 
 	"github.com/babylonchain/babylon/x/btccheckpoint"
 	btccheckpointkeeper "github.com/babylonchain/babylon/x/btccheckpoint/keeper"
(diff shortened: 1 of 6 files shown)


x/epoching/keeper/staking_functions.go
```diff
@@ -5,6 +5,8 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 	stakingtypes "github.com/cosmos/cosmos-sdk/x/staking/types"
+
+	"github.com/babylonchain/babylon/x/epoching/types"
 )
 
 // CheckMsgCreateValidator performs stateless checks on a given `MsgCreateValidator` message
```
SebastianElvis, line 12:
```suggestion
// CheckMsgCreateValidator performs checks on a given `MsgCreateValidator` message
```

Now this function does not only perform stateless checks anymore

x/epoching/keeper/staking_functions.go
```diff
@@ -89,9 +91,15 @@ func (k Keeper) CheckMsgCreateValidator(ctx sdk.Context, msg *stakingtypes.MsgCr
...
 	// sanity check on delegator address
-	if _, err := sdk.AccAddressFromBech32(msg.DelegatorAddress); err != nil {
+	delegatorAddr, err := sdk.AccAddressFromBech32(msg.DelegatorAddress)
+	if err != nil {
 		return err
 	}
 
+	balance := k.bk.GetBalance(ctx, delegatorAddr, msg.Value.GetDenom())
+	if msg.Value.IsGTE(balance) {
+		return types.ErrInsufficientBalance
```
SebastianElvis, line 101: Is there any error that we can reuse from `x/bank` module?

app/app.go
```diff
@@ -346,7 +350,7 @@ func NewBabylonApp(
 
 	// NOTE: the epoching module has to be set before the chekpointing module, as the checkpointing module will have access to the epoching module
 	epochingKeeper := epochingkeeper.NewKeeper(
-		appCodec, keys[epochingtypes.StoreKey], keys[epochingtypes.StoreKey], app.GetSubspace(epochingtypes.ModuleName), &app.StakingKeeper,
+		appCodec, keys[epochingtypes.StoreKey], keys[epochingtypes.StoreKey], app.GetSubspace(epochingtypes.ModuleName), app.BankKeeper, &app.StakingKeeper,
```
SebastianElvis, line 353: Should this be
```suggestion
		appCodec, keys[epochingtypes.StoreKey], keys[epochingtypes.StoreKey], app.GetSubspace(epochingtypes.ModuleName), &app.BankKeeper, &app.StakingKeeper,
```

?

x/checkpointing/keeper/msg_server_test.go
```diff
@@ -97,6 +98,43 @@ func FuzzWrappedCreateValidator_InsufficientTokens(f *testing.F) {
 	})
 }
 
+// FuzzWrappedCreateValidator_InsufficientBalance tests adding a new validator
+// but the delegator has insufficient balance to perform delegating
+func FuzzWrappedCreateValidator_InsufficientBalance(f *testing.F) {
```
SebastianElvis, line 103: Another thing we can do is to move tests on `MsgCreateValidator` from Cosmos SDK to here. Wdyt?

SebastianElvis (approved):
Lgtm! Just some minor comments

x/epoching/keeper/staking_functions.go
gitferry: No. I suppose it's up to the user of the bank module to determine whether the balance is sufficient or not

app/app.go
gitferry: I was wondering the same thing that why the staking keeper is passed as a pointer but not the bank keeper, but it turned out it should be as it is here...

x/checkpointing/keeper/msg_server_test.go
gitferry: I definitely agree but I couldn't find any tests in cosmos-sdk that do stateful tests like us here

vitsalis (approved):
Nice work!

PR #320, opened 2023-02-23
Use cosmos v47

KonradStaniec:
Bump cosmos sdk to 47, ibc-go to v7, and btcd to latest version.

One caveat, for now I disabled e2e test as it seems hermes relayer is not ready for new tendermint version. Long story short, hermes uses events associated with transaction to declare in if some operation succeeded. Format of events was updated with new tendermint version. One easy wasy to confirm that is to run manually some operation on hermes (like `create client` ) and then check `hermes query tx events hash` and and `./babylond query tx hash` , even though babylon node shows all events are there, hermes does not recoignize them.

UPDATE: hermes 1.3 actually returns an error when trying to retrieve events by cli:
`ERROR relayer error: RPC error to endpoint http://bbn-test-a-node-babylon-default-a-3:26657/: serde parse error: incomplete utf-8 byte sequence from index 1 at line 22 column 30`

We can re-enable those after support will be added.


cmd/babylond/cmd/genesis.go
```diff
@@ -146,7 +146,8 @@ func PrepareGenesis(clientCtx client.Context, appState map[string]json.RawMessag
 
 	// gov module genesis
 	govGenState := govv1.DefaultGenesisState()
-	govGenState.DepositParams = &genesisParams.GovParams.DepositParams
+	govGenState.Params.MinDeposit = genesisParams.GovParams.MinDeposit
+	govGenState.Params.MaxDepositPeriod = genesisParams.GovParams.MaxDepositPeriod
```
vitsalis, line 150: Why don't we make `govGenState.Params = genesisParams.GovParams`?

vitsalis: For example, below, we only set the `MinDeposit`, while all the other ones are the defaults.

cmd/babylond/cmd/testnet.go
```diff
@@ -165,6 +165,7 @@ func InitTestnet(
 	babylonConfig.BtcConfig.CheckpointTag = btcCheckpointTag
 	// Explorer related config. Allow CORS connections.
 	babylonConfig.API.EnableUnsafeCORS = true
+	babylonConfig.GRPC.Address = "0.0.0.0:9090"
```
vitsalis, line 168: Isn't this set to this address by default?

go.mod
```diff
@@ -3,73 +3,72 @@ go 1.19
 module github.com/babylonchain/babylon
 
 require (
-	github.com/btcsuite/btcd v0.22.3
-	github.com/cosmos/cosmos-sdk v0.46.6
+	github.com/btcsuite/btcd v0.23.4
```
vitsalis, line 6: Finally 🚀 👏

simapp/utils.go
```diff
@@ -30,7 +29,7 @@ func SetupSimulation(dirPrefix, dbName string) (simtypes.Config, dbm.DB, string,
 	}
 
 	config := NewConfigFromFlags()
-	config.ChainID = helpers.SimAppChainID
+	config.ChainID = "simulation-app"
```
vitsalis, line 32: Use `params.SimAppChainId`?

x/btccheckpoint/module_simulation.go
```diff
@@ -34,14 +34,14 @@ func (AppModule) GenerateGenesisState(simState *module.SimulationState) {
...
 
 // ProposalContents doesn't return any content functions for governance proposals
-func (AppModule) ProposalContents(_ module.SimulationState) []simtypes.WeightedProposalContent {
+func (AppModule) ProposalContents(_ module.SimulationState) []simtypes.WeightedProposalMsg {
 	return nil
 }
 
 // RandomizedParams creates randomized  param changes for the simulator
-func (am AppModule) RandomizedParams(_ *rand.Rand) []simtypes.ParamChange {
+func (am AppModule) RandomizedParams(_ *rand.Rand) []simtypes.LegacyParamChange {
```
vitsalis, line 42: Is there a non-legacy attribute?

x/btclightclient/keeper/state.go
```diff
@@ -230,7 +234,8 @@ func (s headersState) GetMainChain() []*types.BTCHeaderInfo {
 }
 
 // GetHighestCommonAncestor traverses the ancestors of both headers
-//  						to identify the common ancestor with the highest height
+//
+//	to identify the common ancestor with the highest height
```
vitsalis, line 238: Empty lines added to the docstrings of the above functions.

x/btclightclient/module.go
```diff
@@ -122,19 +122,9 @@ func (am AppModule) Name() string {
...
-	return sdk.NewRoute(types.RouterKey, NewHandler(am.keeper))
-}
-
 // QuerierRoute returns the capability module's query routing key.
 func (AppModule) QuerierRoute() string { return types.QuerierRoute }
 
-// LegacyQuerierHandler returns the capability module's Querier.
-func (am AppModule) LegacyQuerierHandler(legacyQuerierCdc *codec.LegacyAmino) sdk.Querier {
-	return nil
-}
```
vitsalis, line 136: Why do we remove the above methods?

x/epoching/keeper/epoch_msg_queue_test.go
```diff
@@ -200,11 +200,12 @@ func FuzzHandleQueuedMsg_MsgWrappedUndelegate(f *testing.F) {
...
 		unbondingDels := helper.StakingKeeper.GetAllUnbondingDelegations(ctx, genAddr)
-		require.Equal(t, 1, len(unbondingDels))                              // there is only 1 type of tokens
-		require.Equal(t, numNewUndels, int64(len(unbondingDels[0].Entries))) // there are numNewUndels entries
-		for _, entry := range unbondingDels[0].Entries {
-			require.Equal(t, coinWithOnePower.Amount, entry.Balance) // each unbonding delegation entry has tokens of 1 voting power
-		}
+		require.Equal(t, 1, len(unbondingDels)) // there is only 1 type of tokens
+
+		// from cosmos v47, all undelegations made at the same height are represented
+		// by one entry see: https://github.com/cosmos/cosmos-sdk/pull/12967
```
vitsalis, line 206: Good, this might help with the issue that we were facing with multiple `UNBONDING` entries for one `UNBONDED` entry cc @SebastianElvis . However, as we execute the wrapped messages, we do not know if more will be made in a next message on the same block. Maybe we could try replicating the above behavior with our messages to avoid any unintended consequences?

x/epoching/keeper/hooks.go
```diff
@@ -134,3 +134,7 @@ func (h Hooks) AfterRawCheckpointForgotten(ctx sdk.Context, ckpt *checkpointingt
 func (h Hooks) AfterRawCheckpointBlsSigVerified(ctx sdk.Context, ckpt *checkpointingtypes.RawCheckpoint) error {
 	return nil
 }
+
+func (h Hooks) AfterUnbondingInitiated(ctx sdk.Context, id uint64) error {
```
vitsalis, line 138: This method does not seem to be used anywhere.

Makefile
```diff
@@ -351,18 +351,15 @@ devdoc-update:
...
-containerProtoImage=tendermintdev/sdk-proto-gen:$(containerProtoVer)
-containerProtoGen=babylon-sdk-proto-gen-$(containerProtoVer)
-containerProtoGenSwagger=cosmos-sdk-proto-gen-swagger-$(containerProtoVer)
-containerProtoFmt=babylon-sdk-proto-fmt-$(containerProtoVer)
+protoVer=0.11.5
+protoImageName=ghcr.io/cosmos/proto-builder:$(protoVer)
+protoImage=$(DOCKER) run --rm -v $(CURDIR):/workspace --workdir /workspace $(protoImageName)
 
-proto-all: proto-gen proto-swagger-gen
+proto-all: proto-format proto-lint proto-gen
```
SebastianElvis, line 358: Why do we ditch `proto-swagger-gen` here?

x/epoching/keeper/epoch_msg_queue_test.go
SebastianElvis: You mean merging multiple unbonding requests into a single one? I feel this is kind of no go. Suppose a scenario where a guy submits two unbonding requests, where the first can succeed while the second will not due to the out-of-balance issue. In this case the epoching module is expected to handle the first one while rejecting the second one. If we merge them together both will fail.

x/epoching/keeper/hooks.go
SebastianElvis: This is because the epoching module's hook has to implement all functions requested by the hook interface of checkpointing. Simply returning a nil means this hook function does nothing.

SebastianElvis (approved):
Finally we can get rid of the btcd issue! I only have some minor comments. Maybe we can wait for followup comments from @vitsalis before merging this

app/app_test.go
```diff
@@ -68,6 +78,8 @@ func TestUpgradeStateOnGenesis(t *testing.T) {
 	ctx := app.NewContext(false, tmproto.Header{})
 	vm := app.UpgradeKeeper.GetModuleVersionMap(ctx)
 	for v, i := range app.mm.Modules {
-		require.Equal(t, vm[v], i.ConsensusVersion())
+		if i, ok := i.(module.HasConsensusVersion); ok {
```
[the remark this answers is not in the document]
KonradStaniec: then we do not need to test it here. From what I understand , this test mainly checks that upgrade keeper versions are inline with versions declared by modules themselves

app/export.go
```diff
@@ -16,8 +16,9 @@ import (
 // ExportAppStateAndValidators exports the state of the application for a genesis
 // file.
 func (app *BabylonApp) ExportAppStateAndValidators(
-	forZeroHeight bool, jailAllowedAddrs []string,
-) (servertypes.ExportedApp, error) {
+	forZeroHeight bool,
+	jailAllowedAddrs []string,
+	modulesToExport []string) (servertypes.ExportedApp, error) {
```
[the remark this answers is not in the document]
KonradStaniec: good catch! we should use here new method for exporting genesis i.e have `app.mm.ExportGenesisForModules(ctx, app.appCodec, modulesToExport)` instead of `app.mm.ExportGenesis(ctx, app.appCodec)`

cmd/babylond/cmd/add_gen_bls.go
```diff
@@ -15,7 +16,7 @@ import (
 	"github.com/babylonchain/babylon/x/checkpointing/types"
 )
 
-func AddGenBlsCmd() *cobra.Command {
+func AddGenBlsCmd(validator genutiltypes.MessageValidator) *cobra.Command {
```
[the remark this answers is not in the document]
KonradStaniec: it is needed as `genutiltypes.ValidateAndGetGenTx` require new argument of type `genutiltypes.MessageValidator`. This argument should be consistent across whole app and all commands so we need to pass it from `root.go` to all necessary call sites.  

From what I understand, this new argument makes it possible to customize validator message which is put in genesis file. (at least validations done on that message) So it may come in handy at some point, for example to require custom create validator message with bls key. For now everywhere I am passing just the default implementation `DefaultMessageValidator`

cmd/babylond/cmd/testnet.go
KonradStaniec: it was, but it was changed to `localhost` which kinda make sense from security perspective i.e now it must be conscious decision to open the node to whole world.

x/btccheckpoint/module_simulation.go
KonradStaniec: there is not, altthough while looking at it I noticed that `RandomizedParams ` function was ditched by cosmos-sdk and removed from simpulation interface, therefore I think I wil lust delete that function from all our mnodules (and therefore delete LegacyParamChange from everywhere)

x/btclightclient/module.go
KonradStaniec: those function and types were removed from cosmos-sdk - https://github.com/cosmos/cosmos-sdk/blob/release/v0.47.x/UPGRADING.md#appmodule-interface

Makefile
KonradStaniec: good point, I just got carried away and copied stuff from cosmos-sdk. (we do not even have proto-format and proto-lint)

I will bring it back to how it was and crated task to and linter and formatter for protos- https://github.com/babylonchain/babylon/issues/321

vitsalis (approved):
Awesome, thanks for the updates!

cmd/babylond/cmd/add_gen_bls.go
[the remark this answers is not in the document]
gitferry: This is a nice feature introduced by `v47`, which allows us to customize the validation rules against `staking.MsgCreateValidator` messages. I think for now the `DefaultMessageValidator` is sufficient, which solely calls the `ValidateBasic` method of the `staking.MsgCreateValidator`. I don't think we should add more rules about BLS keys because BLS keys are added after genesis txs are generated.

PR #322, opened 2023-03-03
Update protobuf generation

KonradStaniec:
Various cleanups in our protobufs:
- use cosmos protobuf fork - https://github.com/cosmos/cosmos-sdk/blob/release/v0.47.x/UPGRADING.md#protobuf
- delete some unused `x/epoching/types/message.pb.go` proto file which was with us from the start of the project
- add proto-lint and proto-format commands
- fix linter errors. Most important ones fixed is a lot of unused imports and moving all protos to correct folders i.e protos in v1 package should be in v1 folder.

I have also for now turned off a lot of linter rules regarding comments in proto files, as our protos are definitly under documented 😅 

Rules I am talking about are:
```
    # TODO Decide which comments we would like to enfore by linter
    - COMMENT_MESSAGE
    - COMMENT_ENUM_VALUE
    - COMMENT_ENUM
    - COMMENT_RPC
    - COMMENT_ONEOF
```

Imo we should decide which of those rules we want to enforce , and then everybody should pick one module and add proper docs, then we can re-enable this rules.


Makefile
```diff
@@ -8,7 +8,6 @@ BINDIR ?= $(GOPATH)/bin
 BUILDDIR ?= $(CURDIR)/build
 HTTPS_GIT := https://github.com/babylonchain/babylon.git
 DOCKER := $(shell which docker)
-DOCKER_BUF := $(DOCKER) run --rm -v $(CURDIR):/workspace --workdir /workspace bufbuild/buf
```
SebastianElvis, line 11: Why do we remove this?

proto/babylon/btccheckpoint/v1/btccheckpoint.proto
```diff
@@ -43,37 +43,35 @@ message BTCSpvProof {
 // which transaction was included and transaction index in the block
 message TransactionKey {
   uint32 index = 1;
-  bytes hash = 2 [
-    (gogoproto.customtype) = "github.com/babylonchain/babylon/types.BTCHeaderHashBytes"
```
SebastianElvis, line 47: If I understand correctly, this format is enforced by `proto-lint`, right?

SebastianElvis (approved):
Thanks! Finally we got rid of the protobuf error.

Regarding the lint rules, maybe each of us can create tickets for modules which one is responsible for. Wdyt?

Makefile
KonradStaniec: it was not used anywhere, also all `buf` usage comes from `protoImage`

proto/babylon/btccheckpoint/v1/btccheckpoint.proto
KonradStaniec: this is mostly from `proto-format`. Currently it uses default config but if we will find some formatting weird we will be able to adjust it it in the future

KonradStaniec:
> Regarding the lint rules, maybe each of us can create tickets for modules which one is responsible for. Wdyt?

Sounds like good idea, maybe we can disscus it on our next sprint planning meeting.

PR #326, opened 2023-03-15
Call custom babylon epochin module from smart contract  [DO NOT MERGR]

KonradStaniec:
Just minimal example how to call custom babylon module from smart contract


contracts/babylon-bindings/src/querier.rs
```diff
@@ -0,0 +1,19 @@
+use cosmwasm_std::{StdResult, Uint64, QuerierWrapper};
+
+use crate::query::{BabylonQuery, CurrentEpochResponse};
+
+pub struct BabylonQuerier<'a> {
+  querier: &'a QuerierWrapper<'a, BabylonQuery>,
```
SebastianElvis, line 6: I'm awful in Rust so a noob question 😅 What does `'a` mean in these three places and why are they needed? Does it mean `BabylonQuery` objects and `QuerierWrapper` objects always share the same lifecycle or something else? Also why couldn't we use `BabylonQuery` objects directly?

KonradStaniec: so from I understand, it means that `BabylonQuerier ` struct will live as long as `QuerierWrapper`  inside which actually makes sense because `QuerierWrapper` is passed as a reference here.

We could use BabylonQuery directly in our smart contracts, but from I understand this is kind of convention which just makes it easier for other smart contracts to call chain custom modules. (here is smilar bindings lib in osmosis - https://github.com/osmosis-labs/bindings/blob/main/packages/bindings/src/querier.rs)

KonradStaniec:
closing this up as we already better way of integrating custom queries - https://github.com/babylonchain/babylon/pull/347

PR #332, opened 2023-03-22
Use new version for our fork of ibc go

KonradStaniec:
New version has been simplified due to usage of ibc-go v7 
New version can work with other types of light clients not only tendermint

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -203,7 +203,7 @@ replace (
 
 	// slay the dragonberry
 	github.com/confio/ics23/go => github.com/cosmos/cosmos-sdk/ics23/go v0.8.0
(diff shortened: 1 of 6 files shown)


SebastianElvis (approved):
Lgtm!

A random idea: is it feasible to move the `ExtendedKeeper` from IBC-Go fork repo to Babylon? such that the modifications in IBC-Go fork are made minimal and we can update stuff related to this hook within a single PR

KonradStaniec:
> A random idea: is it feasible to move the ExtendedKeeper from IBC-Go fork repo to Babylon? such that the modifications in IBC-Go fork are made minimal and we can update stuff related to this hook within a single PR

yup I was actually thinking about it. 

I think the only thing that needs to live in fork repo is introducing keeper interface and using it in tests and necessary places. ExtendedKeeper implementation and hooks can be moved to our babylon repo.

PR #333, opened 2023-03-23
zoneconcierge: moving extended client keeper to Babylon repo

SebastianElvis:
Context: https://github.com/babylonchain/babylon/pull/332#issuecomment-1480688869

This PR moves the `ExtendedKeeper` from IBC-Go fork repo to Babylon. This simplifies the maintenance of IBC-Go fork: the only difference between our IBC-Go fork and original IBC-Go will be the client keeper interface.


KonradStaniec (approved):
lgtm!

Just question, maybe it makes sense to remove this stuff on ibc-go repo side  and update ibc-go fork version in this pr ? (this way we make 100% sure we are using babylon local structs and methods)

Issue #325, opened 2023-03-13
Remove usage of x/params module


KonradStaniec:
cosmos-sdk module `x/params` have beed depraceted and will be removed- https://github.com/cosmos/cosmos-sdk/blob/main/UPGRADING.md#xparams. 

All Babylon modules should move to new way of managing params, and we should remove params usage from the app.

PR #334, opened 2023-03-23
Cleanup params in Babylon custom modules.

KonradStaniec:
A bit of preparation before fixing - https://github.com/babylonchain/babylon/issues/325

A lot of babylon modules used empty `Params` protos. This makes little sense as it just generate a lot unnecessary code. This pr removes those `Params` protos and all related code.


proto/babylon/checkpointing/v1/genesis.proto
```diff
@@ -3,18 +3,14 @@ package babylon.checkpointing.v1;
...
 import "cosmos/crypto/ed25519/keys.proto";
-import "babylon/checkpointing/v1/params.proto";
 import "babylon/checkpointing/v1/bls_key.proto";
 
 option go_package = "github.com/babylonchain/babylon/x/checkpointing/types";
 
 // GenesisState defines the checkpointing module's genesis state.
 message GenesisState {
-  // params defines all the paramaters of related to checkpointing
-  Params params = 1 [ (gogoproto.nullable) = false ];
```
SebastianElvis, line 14: It it worth keeping the `Params` message and field here in case of any new parameters in the future?

proto/babylon/btclightclient/v1/genesis.proto
```diff
@@ -2,14 +2,11 @@ syntax = "proto3";
...
 
 import "gogoproto/gogo.proto";
-import "babylon/btclightclient/v1/params.proto";
 import "babylon/btclightclient/v1/btclightclient.proto";
 
 option go_package = "github.com/babylonchain/babylon/x/btclightclient/types";
 
 // GenesisState defines the btclightclient module's genesis state.
 message GenesisState {
-  Params params = 1 [ (gogoproto.nullable) = false ];
```
SebastianElvis, line 12: same here, is it worth keeping params?

proto/babylon/monitor/v1/genesis.proto
```diff
@@ -2,9 +2,8 @@ syntax = "proto3";
 package babylon.monitor.v1;
 
 import "gogoproto/gogo.proto";
-import "babylon/monitor/v1/params.proto";
 
 option go_package = "github.com/babylonchain/babylon/x/monitor/types";
 
 // GenesisState defines the monitor module's genesis state.
-message GenesisState { Params params = 1 [ (gogoproto.nullable) = false ]; }
```
SebastianElvis, line 10: same here

proto/babylon/zoneconcierge/v1/genesis.proto
```diff
@@ -2,12 +2,10 @@ syntax = "proto3";
 package babylon.zoneconcierge.v1;
 
 import "gogoproto/gogo.proto";
-import "babylon/zoneconcierge/v1/params.proto";
 
 option go_package = "github.com/babylonchain/babylon/x/zoneconcierge/types";
 
 // GenesisState defines the zoneconcierge module's genesis state.
 message GenesisState {
-  Params params = 1 [ (gogoproto.nullable) = false ];
```
SebastianElvis, line 11: Same here. Notably, a param that can be upgraded by gov prop might be useful when we add support to phase 3 when at phase 2.

SebastianElvis (commented):
Thanks for the effort and this indeed cuts a lot of code. However, I have some concern that removing boilerplate code for params might make it hard for us to add new parameters to these modules in the future. It could be possible that we want some new parameters in a certain module, and may even want to update their values upon gov prop. Given these reasons I'm inclined to keep the code as is. No strong objection, but would like to hear feedback from more ppl

vitsalis (commented):
Agree with @SebastianElvis here. Do we get anything else other than less boilerplate code if we remove them? From what I understand from #325 , we are already compliant with the change (at least with the changes related to this PR), as we store module parameters in the module.

KonradStaniec:
> Do we get anything else other than less boilerplate code if we remove them? From what I understand from https://github.com/babylonchain/babylon/issues/325 , we are already compliant with the change (at least with the changes related to this PR), as we store module parameters in the module.

So currently we store params in x/params module which is now deprecated and from cosmos sdk 0.48 will be removed. Given the fact that we are currently always on newest version, I assume we will also want to switch to 0.48, so we will want to make a switch to a new version. It makes sense to do it before mainnet as this way we will avoid any upgrade handlers etc.

> however, I have some concern that removing boilerplate code for params might make it hard for us to add new parameters to these modules in the future. It could be possible that we want some new parameters in a certain module, and may even want to update their values upon gov prop. Given these reasons I'm inclined to keep the code as is. No strong objection, but would like to hear feedback from more ppl

So the effort to add params later from scratch versus add new param to params proto will be roughly the same. You will need upgrade handler which will update required proto files or create v2 of proto file, and probably gov proposal to introduce them.

Also current situation is mostly a result of copy/pasting modules at the beginning of our project which is not really good design.

In general carrying code which may/or may not become useful in some not specified time in the future is a smell that something is not right. 

In general, we have still some time till mainnet, so we have time to decide if any of those modules will require params. If we happen to decide that this is the case, it will be done in new way which is much different from current one. 

So in summary imo it makes sense to:
- start from deleting old way of handling params from modules which does not need it as a) modules do not use params eitherway, b) current way of handling params is deprecated and will be removed
- move remaining modules to new way of handling params.
- decide which of other modules will need params, and if any ,and implement it in new way. 

Keeping 4k lines of code just in case just sounds like a bad deal.

Issue #336, opened 2023-03-25
Remove `handler.go` from modules


fadeev:
Starting from version 0.46 of the Cosmos SDK, defining message handling [is no longer necessary](https://github.com/cosmos/cosmos-sdk/pull/9650). The `NewHandler()` function in `handler.go` is currently unused in the code, making `handler.go` files safe to remove.

PR #337, opened 2023-03-25
refactor: remove `handler.go` files from modules

fadeev:
`handler.go` files are no longer necessary. Modified test files to use `MsgSrvr`, instead.

close https://github.com/babylonchain/babylon/issues/336


vitsalis (approved):
Nice, thanks for the catch!

PR #338, opened 2023-03-27
Enhancement/grandpa parser

edjroz:
Provide an implementation of the scale codec to encode & decode header hashes


KonradStaniec:
Only general comments:
1:  Lets target branch - https://github.com/babylonchain/babylon/tree/centauri-integration-ibc-v7-stable
2. Maybe we can create directory on path `x/zoneconcierge/types/grandpa` and move the whole logic there ? `zoneconcierge` is our main module responsible for checkpointing and I feel weird having this this it root of the repo 😅

PR #340, opened 2023-03-28
docker: Refactor to a lightweight unified image

danbryan:
Lets chat about this in our next sync.  I don't think we need to call out things like EXPOSE, and  ENTRYPOINT in the dockerfile.  Alot of that can be defined in the pod spec when we call this image.


danbryan:
@vitsalis and @philmln lets chat about this too, for the time being i have hard coded the build arguments into the Dockerfile.  But ideally, circle CI should be passing these into the Dockerfile

vitsalis:
Updated the docker-compose and Makefile to properly handle the new docker image by executing a full command. This previously failed due to the `docker-compose` file assuming that there will be a `wrapper.sh` script.

Further, I added the conditional addition of the `dlv` debug package based on a debug flag being set. Unfortunately, I could only do the conditional copying of the `dlv` binary through a hacky mechanism. I did this so that we don't have to maintain two separate docker images just for having the dlv package available. Further, the `contrib/images/babylon-dlv/wrapper.sh` script contains some dlv specific commands that should be added to the `docker-compose` file. However, this might lead to much extra unneeded code there. @KonradStaniec maybe we can remove the support for a `dlv` docker image, given that I don't think anybody is using it right now.

vitsalis:
@danbryan currently the build fails due to a permission issue. Seems that we're not able to create a directory inside the container due to permissions. I tried many ways to resolve this but with no success. Any thoughts?

KonradStaniec:
> Overall, I have not ever used the dlv docker image and I have my doubts about anyone else using it. In this repo we should maintain a minimal and production ready Docker image. If we want to have dlv we can introduce this functionality, but we can also potentially remove it to avoid having to manage this extra overhead. @KonradStaniec what do you think? Should we keep dlv support?

tbh I also never used it, but I do not know about other people. Maybe we can bring it up on next eng meeting ?

contrib/images/babylond/Dockerfile
```diff
@@ -0,0 +1,127 @@
...
+# Build
+WORKDIR /go/src/github.com/babylonchain
+RUN git clone https://github.com/babylonchain/babylon.git
+WORKDIR /go/src/github.com/babylonchain/babylon
+RUN git fetch
+RUN git checkout ${VERSION}
+WORKDIR /go/src/github.com/babylonchain/babylon
+
+RUN if [ "${TARGETPLATFORM}" = "linux/amd64" ] ; then \
+      wget -P /lib https://github.com/CosmWasm/wasmvm/releases/download/v1.1.1/libwasmvm_muslc.x86_64.a ; \
```
KonradStaniec, line 32: I would say hardcoding wasmvm version is definitly not good enough here.  On dev branch current docker file contains script to download version used in go.mod and with correct platform, so it can be probably re-used here.

danbryan: good idea, this can definitely be updated.

vitsalis: fixed

vitsalis:
Managed to resolve the issue. The gift of it is:
- Inside the CircleCI environment, there wasn't a `.testnets` file. This means that the docker daemon created the directory to mount it as a volume.
- The directory is created with access rights corresponding to the `circleci` user and group, with IDs `1001` and `1002` respectively. The access permissions are `755`, meaning that only the user and group can write to the volume.
- When the directory gets mounted, it is assigned to the same user and group IDs as well as the same access permissions. This means that the `babylon` user that the docker image uses, does not have write permissions to the `.testnets` directory and hence the permissions error.
- To resolve this, we need to make the `.testnets` directory writeable by entities outside of the group. Hence a solution (which is a bit hacky) is to `chmod 777` the directory. Given that this directory is only used for tests I would say this is acceptable. So this PR does that.

Overall, the PR was reworked a bit from its initial version. More specifically:
1. Introduce a lightweight Docker image that only installs the required binaries to the final image. This deviates a bunch from the Docker image we already used that was copied from the cosmos-sdk repository. This change is needed, as the initial Docker image has been developed to be used as an image for the `localnet-start` command and therefore is a very simple one without optimizations. This is evident by the `wrapper.sh` script that was used by it to wrap commands`. We should have a single Babylon image that can be used for both debugging and production. This will allow us to avoid bugs compared to different builds.
2. Along the image that was used for the localnet, this repository also maintained a Docker image which also included the `dlv` debugging package. We potentially want to include support for it, but do not want to maintain two separate Docker images that include it. For that purpose, this PR introduces (1) a mechanism that conditionally includes the `dlv` package based on a build argument and (2) a `docker-compose.yml.debug` file that spins up a network executing the `babylond` binary wrapped by `dlv`. For (1), the mechanism is a bit hacky as Docker does not allow for conditional `COPY` commands. For (2), we introduce a copy of the `docker-compose.yml` file and only change the command that is executed for the container. We have 3 options: (a) keep support for `dlv`, i.e. what happens in this PR, (b) remove the `docker-compose.yml.debug` file to avoid managing it but keep Docker conditional support for `dlv`, or (c) remove `dlv` support altogether. If nobody is using `dlv`, I suggest we remove it entirely.
3. Rework the structure of the `Makefile`. Some stuff didn't make sense to me but would appreciate a look.

side note: @danbryan the initial PR had a hardcoded version of `v0.5.0`. I suppose that's left over from copying it from the infra repository. I changed this to `HEAD` to always use the latest version and the version to be used to build the Docker image can be specified using the `VERSION` build argument as before.

Please check this out @KonradStaniec @SebastianElvis @danbryan . Would appreciate your thoughts especially on change (2).

SebastianElvis:
> tbh I also never used it, but I do not know about other people. Maybe we can bring it up on next eng meeting ?

Same, I've never used the dlv image.

contrib/images/babylond/Dockerfile
```diff
@@ -0,0 +1,124 @@
...
+fi
+
+RUN LEDGER_ENABLED=$LEDGER_ENABLED BUILD_TAGS=$BUILD_TAGS COSMOS_BUILD_OPTIONS=$COSMOS_BUILD_OPTIONS make install
+
+# Use busybox to create a user
+FROM busybox:stable-musl AS busybox
+RUN addgroup --gid 1137 -S babylon && adduser --uid 1137 -S babylon -G babylon
+
+# Use scratch for the final image
+FROM scratch
```
SebastianElvis, line 49: What's the benefit of using `scratch` compared to `alpine`?

contrib/images/babylond/Dockerfile
```diff
@@ -0,0 +1,124 @@
...
+COPY --from=build-env /lib/libcrypto.so.1.1 /lib
+COPY --from=build-env /usr/lib/libbrotlidec.so.1 /lib
+COPY --from=build-env /usr/lib/libbrotlicommon.so.1 /lib
+
+# Install trusted CA certificates for curl
+COPY --from=build-env /etc/ssl/cert.pem /etc/ssl/cert.pem
+
+# Install cli tools from busybox
+COPY --from=busybox /bin/ln /bin/ln
+COPY --from=busybox /bin/ln /bin/sh
```
SebastianElvis, line 84: It might be possible that the binaries copied from `golang-alpine` are incompatible with `scratch` due to various reasons, e.g., different base OS, different arch, using libraries that do not exist in `scratch`. Thus it's not easy to maintain a list of dynamically linked libraries and binaries in the final image.

Alternatively, could we keep using `alpine` for the final image? It's also minimal, supports adding new binary tools via `apk add`, and has better compatibility with `golang-alpine`?

SebastianElvis (commented):
Great work! I'm fully supportive to the idea of merging `dlv` and `env` to a single image. The only small part I'm not sure is the choice of base image.

contrib/images/babylond/Dockerfile
danbryan: One of the advantages are the images are very small, which leads to quick restarts for pods. You can also remove unwanted binaries, like maybe your concerned about data exfil of the nodekey.json  you can remove apps like cat, tee, cp, scp  anything that could be used  to display the key.  The only binaries are the ones you white list.  I do agree that it does require more maintenance.  I think alpine is just fine too.

contrib/images/babylond/Dockerfile
danbryan: yeah, your right, you have to stay on top of the library updates, but from my experience its not very often you need to do this.   Especially if version control the base  image as we have done here `golang:1.19-alpine3.16` i don't think we would need to worry about them changing until we switch to something like `golang:1.19-alpine3.17` and that would require a one time update most likely.

SebastianElvis: As long as this does not need much effort for maintenance, then lgtm!

SebastianElvis (approved):
Lgtm, great work!

contrib/images/babylond/Dockerfile
KonradStaniec: hmm I did a small test to compare image size on this branch and current dev and this size difference is the other way around i.e
this branch - `babylonchain/babylond           latest    f86350e061f3  316MB`
dev branch - `babylonchain/babylond           latest    e3d235add639  91.2MB`

Is this the result of adding some additional tools to the image ? how large would be the alpine image with those tools ? 

As this approach generates more maintenance effort, It would be really nice to be sure that result images are really visibly smaller than the alpine based ones.

danbryan: This is odd,  i agree it reports being larger than it appears inside.
```
f1c4786c5333   babylonchain/babylond   "/bin/sh"   5 minutes ago   Up 5 minutes             awesome_elion   0B (virtual 332MB)
```

but inside, 

```
docker run -it -u 0:0 --entrypoint /bin/sh docker.io/babylonchain/babylond
```

```
/home/babylon # cd /
/ # du -ckhs * | sort -n
0       dev
0       home
0       proc
0       sys
5.1M    lib
83.1M   bin
232.0K  etc
88.4M   total
```
and if you look at all the files in bin

```
/bin # du -ckhs * | sort -n
1.1M    cat
1.1M    ln
4.0K    ldd
80.4M   babylond
244.0K  curl
272.0K  jq
83.1M   total
```

I'll keep digging, but off the top of my head i'm not sure why it's saying that it's 4x larger than it looks inside.

danbryan: Found it,  this section is causing the inflated size

```
# HACK: Only copy dlv if it exists
COPY --from=build-env /go/bin/ /tmpbin
RUN cp /tmpbin/dlv /bin 2>/dev/null; rm -rf /tmpbin
```
with it, the image is 320MB
without it the image is 90MB

Ill give it a rework to see if i can make the non dlv image smaller

danbryan: Ok, this solves the issue.

https://github.com/babylonchain/babylon/pull/340/commits/14f6e8991148f58ad447da13413532ee28f8782c

The image is now ~90M normal and 110M with dlv

The issue was 2 additional binaries `chain` and `node` were getting copied in somehow.  I couldn't see the inside the file system, but they were making into the layer somehow.  I modified the logic to remove these binaries during the build phase, and just only copy the remaining images, which would be `babylond` and optionally `dlv`

danbryan: done
https://github.com/babylonchain/babylon/pull/340/commits/cdadf079c6ade3dfbaed3df34077e3654d6fd4dc

vitsalis:
After today's eng meeting we made the following decisions:
1. Remove support for dlv in the Docker image entirely
2. No need to copy specific commands from the build environment to whitelist them. We're ok with having all commands available as the size is not affected that much.

To that end, I made the 3-stage build a 2-stage one based on Alpine, and the second part of the build just copies the Babylon binary and executes based on the `babylon` user. So to sum up, the updates of this PR involve:

1. Rework the image to include parameters related to version and build environment among others.
2. Introduce the usage of a Babylon user
3. Remove the `entrypoint.sh` functionality which wrapped the Babylon command. Instead we allow the user of the Docker image to use whatever command they want. Further, this makes the Dockerfile usable as a standalone file.
4. Remove support for dlv

philmln (approved):
LGTM!
Restricting what packages will run in the container is good security-wise, but limits us if we want to run additional operations other than the babylond process.

PR #343, opened 2023-04-04
Feature/lint proto ci

KonradStaniec:
In pr which removed not needed params I have broken proto-linting. The only conclusion is that whatever is not part of regular CI will ultimately rot. This pr add linting of protos as CI step. 

One question I have, maybe to someone who knows more about Circle CI, currently `make proto-lint` will always pull image from ghcr with correct version of `buf`  (which ultimately does the linting), is it possible to cache this image between CI runs to avoid pulling it every time ?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,9 +30,6 @@ jobs:
           paths:
             - "/home/circleci/.go_workspace/pkg/mod"
       - run:
-          name: Lint proto files
-          command: make proto-lint
-      - run:
           name: Run tests
           command: |
             make test
@@ -54,6 +51,9 @@ jobs:
       resource_class: medium
     steps:
       - checkout
+      - run:
+          name: Lint proto files
+          command: make proto-lint
       - run:
           name: Lint
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,6 +30,9 @@ jobs:
           paths:
             - "/home/circleci/.go_workspace/pkg/mod"
       - run:
+          name: Lint proto files
+          command: make proto-lint
+      - run:
           name: Run tests
           command: |
             make test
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -29,6 +29,9 @@ jobs:
           key: go-mod-v6-{{ checksum "go.sum" }}
           paths:
             - "/home/circleci/.go_workspace/pkg/mod"
+      - run:
```
philmln, line 32: Would it make more sense to include this step under the `lint` CI job?

philmln, line 32: Would it make more sense to include this step under the `lint` job?

philmln (approved):
CircleCI has the concept of [Docker Layer Caching](https://circleci.com/docs/docker-layer-caching/) for images that are built during a CI execution. I don't know if it also applies for images that are pulled during execution, but we could give it a shot. Overall though, the time overhead of pulling this specific image is under 10 secs, so it doesn't seem too important at this point.

Docker Layer Caching seems more interesting when we build babylond. It currently takes up to 90 seconds, so we could benefit by reducing some of the overhead.

.circleci/config.yml
KonradStaniec: yup 👍 I forgot that now we have separate `lint` job

KonradStaniec:
> CircleCI has the concept of [Docker Layer Caching](https://circleci.com/docs/docker-layer-caching/) for images that are built during a CI execution. I don't know if it also applies for images that are pulled during execution, but we could give it a shot. Overall though, the time overhead of pulling this specific image is under 10 secs, so it doesn't seem too important at this point.

Docker Layer Caching seems more interesting when we build babylond. It currently takes up to 90 seconds, so we could benefit by reducing some of the overhead.

Hmm sounds like something good to experiment with. In general, I agree pulling this image does not take to much time now, so for now I will merge it as it is, and we can figure out later any improvements to this process.

PR #345, opened 2023-04-08
fix: Make e2e tests work with new docker image

vitsalis:


KonradStaniec (approved):
looks good 👍 

I wonder, if CircleCI, has the option to run CI on pull_request events (similarly to gh actions) and then the workflow is run against base branch i.e what is run is base branch + chanages from pr ? If that is possible then merging such breaking change would be impossible.

PR #347, opened 2023-04-12
Custom bindings for smart contracts

KonradStaniec:
- Adds support for custom query available for smart contract based on https://github.com/babylonchain/bindings. Currently only one query is supported i.e gettich current babylon epoch
- Add e2e test and test setup to check that everything is wired correctly and we are able to pass data between custom modules and wasmd.

One thing we may need talk through is storing wasm blobs in repo, instead of building them with each run. Imo it makes sense for now for the sake of CI speed. As soon as the custom api will be stable, we can release them directly in https://github.com/babylonchain/bindings.


SebastianElvis:
> One thing we may need talk through is storing wasm blobs in repo, instead of building them with each run.

This looks fine from my side, given that 1) the wasm file is small after optimisation, 2) storing it in the repo allows everyone to  use a consistent wasm file, 3) this speeds up all scenarios using wasm file, e.g., CI. As an example, Osmosis also stores its [rate limiter wasm file](https://github.com/osmosis-labs/osmosis/blob/main/x/ibc-rate-limit/bytecode/rate_limiter.wasm) and [wasm files for testing purposes](https://github.com/osmosis-labs/osmosis/blob/main/wasmbinding/testdata/osmo_reflect.wasm) in the code base.

Makefile
```diff
@@ -417,6 +423,17 @@ localnet-stop:
 # localnet-test-integration will spin up a localnet and run integration tests on it
 localnet-test-integration: localnet-start test-integration localnet-stop
 
+build-test-wasm:
+	cargo build --release --lib --target wasm32-unknown-unknown --manifest-path wasmbinding/testdata/Cargo.toml
```
SebastianElvis, line 427: This line may not be necessary since the below Docker container will copy all source code and compile it from scratch anyway

app/app.go
```diff
@@ -150,7 +150,7 @@ const (
 	// The last arguments can contain custom message handlers, and custom query handlers,
 	// if we want to allow any custom callbacks
 	// See https://github.com/CosmWasm/cosmwasm/blob/main/docs/CAPABILITIES-BUILT-IN.md
-	wasmCapabilities = "iterator,staking,stargate,cosmwasm_1_1,cosmwasm_1_2"
+	wasmCapabilities = "iterator,staking,stargate,cosmwasm_1_1,cosmwasm_1_2, babylon"
```
SebastianElvis, line 153:
```suggestion
	wasmCapabilities = "iterator,staking,stargate,cosmwasm_1_1,cosmwasm_1_2,babylon"
```

Not sure if space is allowed here though

wasmbinding/wasm.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+
+		switch {
+		case contractQuery.Epoch != nil:
+			epoch := qp.epochingKeeper.GetEpoch(ctx)
+			res := bindings.CurrentEpochResponse{
+				Epoch: epoch.EpochNumber,
+			}
+
+			ctx.Logger().Info("Marshalled cutom response")
```
SebastianElvis, line 42:
```suggestion
			ctx.Logger().Debug("Marshalled cutom response")
```

SebastianElvis (approved):
Lgtm 👍 Osmosis does provide a very solid and informative reference for us

app/app.go
KonradStaniec: tbh everywhere I looked capabilities were defined without whitespaces, so I would leave it like that 😅

vitsalis: But here there's a whitespace, no? 😄

wasmbinding/wasm.go
```diff
@@ -0,0 +1,69 @@
...
+func NewQueryPlugin(ek *epochingkeeper.Keeper) *QueryPlugin {
+	return &QueryPlugin{
+		epochingKeeper: ek,
+	}
+}
+
+// CustomQuerier dispatches custom CosmWasm bindings queries.
+func CustomQuerier(qp *QueryPlugin) func(ctx sdk.Context, request json.RawMessage) ([]byte, error) {
+	return func(ctx sdk.Context, request json.RawMessage) ([]byte, error) {
+
```
vitsalis, line 29: empty line

vitsalis (approved):
Nice!

app/app.go
KonradStaniec: aah I missed what is diff and what if proposed change 🤦

PR #350, opened 2023-04-17
Add support for querying for latest finalized epoch

KonradStaniec:
Add support for querying what is latest finalized epoch from smart contract side.

This along with current epoch query enables few most basic use cases, for example smart contract which saves arbitrary data and some epoch, and exposes query whether this data is finalized.


wasmbinding/wasm.go
```diff
@@ -9,17 +9,23 @@ import (
 	wasmvmtypes "github.com/CosmWasm/wasmvm/types"
 	"github.com/babylonchain/babylon/wasmbinding/bindings"
 	epochingkeeper "github.com/babylonchain/babylon/x/epoching/keeper"
+	zoneconciergeKeeper "github.com/babylonchain/babylon/x/zoneconcierge/keeper"
```
SebastianElvis, line 12:
```suggestion
	zoneconciergekeeper "github.com/babylonchain/babylon/x/zoneconcierge/keeper"
```

or 

```suggestion
	zckeeper "github.com/babylonchain/babylon/x/zoneconcierge/keeper"
```

x/zoneconcierge/keeper/epochs.go
```diff
@@ -19,7 +19,8 @@ func (k Keeper) GetFinalizedEpoch(ctx sdk.Context) (uint64, error) {
 
 // setFinalizedEpoch sets the last finalised epoch
 // called upon each AfterRawCheckpointFinalized hook invocation
-func (k Keeper) setFinalizedEpoch(ctx sdk.Context, epochNumber uint64) {
+// NOTE: Public for test purposes
+func (k Keeper) SetFinalizedEpoch(ctx sdk.Context, epochNumber uint64) {
```
SebastianElvis, line 23: Instead of making this public, how about use `AfterRawCheckpointFinalized` to indirectly trigger this function for testing purposes?

vitsalis (approved):

SebastianElvis (approved):
Thanks!

wasmbinding/wasm.go
KonradStaniec: like the second one 👍

PR #352, opened 2023-04-20
fix: fix query height in `ProveEpochSealed`

SebastianElvis:
This PR fixes a bug in `ProveEpochSealed`. 

When executing an ABCI query `abci.RequestQuery`, one needs to feed a `height`, such that the node will reply an inclusion proof of the queried KV pair against the header at `height+1`. 

We use this ABCI query for generating proofs of inclusion for the validator set and the metadata of epoch `i` against the sealer header, i.e., the 2nd header of epoch `i+1`. In order to allow verifier to verify the proofs against the sealer header, the ABCI query should be given height `queryHeight-1` rather than `queryHeight`, where `queryHeight` will be the sealer header's height. This PR fixes the bug where we use `queryHeight` rather than `queryHeight - 1`.

The corresponding unit test is also fixed.


vitsalis (approved):
Nice catch!

PR #356, opened 2023-04-21
Move checkpoint tag to params

KonradStaniec:
- moves checkpoint tag to params
- make checkpoint tag hex encoded string at babylon node level. It is job of some high level protocols to give those bytes a meaning.
- removes file with default tags


testutil/datagen/btc_transaction.go
```diff
@@ -375,9 +375,10 @@ func GenRandomBabylonTxPair() ([]*wire.MsgTx, *btctxformatter.RawBtcCheckpoint)
 
 	// fake a raw checkpoint
 	rawBTCCkpt := GetRandomRawBtcCheckpoint()
+	tag := []byte{1, 2, 3, 4}
```
vitsalis, line 378:
```suggestion
	tag := datagen.GenRandomByteArray(4)
```
will resolve the `TODO` below.

vitsalis (approved):
Very clean! As we discussed offline, it would be nice if the BTC network was also moved to params as well, but given that the `btccheckpoint` module also needs to verify the work of headers as it receives them, this is not entirely straightforward. One case in which we could keep this feature (i.e. btccheckpoint receives headers as well) would be that the module doesn't do any PoW checks itself, but instead just forwards the header and the `btclightclient` module performs the check, returning error if it's not valid.

KonradStaniec:
> One case in which we could keep this feature (i.e. btccheckpoint receives headers as well) would be that the module doesn't do any PoW checks itself, but instead just forwards the header and the btclightclient module performs the check, returning error if it's not valid. We can further discuss this as a future optimization

It would be doable, but it would require a bit of refactoring on `btccheckpoint` module part, I am not sure if it won't be better to just receive headers hashes instead of whole headers as we do not use headers as second source eitherway :)

PR #358, opened 2023-04-24
Version bumps

KonradStaniec:
bumps:
- comos-sdk to 0.47.1
- go to 1.20
- wasmd to 0.40.0-rc1

Also little fix:
- remove `staking` capability from smart contracts. Out staking is going through `epoching` module so it seems we should not allow smart contracts to send staking messages directly to `staking` module. There is probably a way to passing `epoching` module to `wasmd` keeper, but it would require a bit investigation so it seems safer to just disable it for now.


x/zoneconcierge/keeper/proof_tx_in_block_test.go
```diff
@@ -54,10 +54,14 @@ func TestProveTxInBlock(t *testing.T) {
...
 	msg := banktypes.NewMsgSend(val.Address, testNetwork.Validators[1].Address, coins)
 
+	txf, err := tx.NewFactoryCLI(val.ClientCtx, fs)
+
+	require.NoError(t, err)
 	// construct a tx that includes this msg
-	txf := tx.NewFactoryCLI(val.ClientCtx, fs).
+	txf = txf.
 		WithTxConfig(val.ClientCtx.TxConfig).
 		WithAccountRetriever(val.ClientCtx.AccountRetriever)
```
SebastianElvis, line 63: Do we need to assign here again, or can simply

```suggestion
	txf.WithTxConfig(val.ClientCtx.TxConfig)
	     .WithAccountRetriever(val.ClientCtx.AccountRetriever)
```

?

x/checkpointing/client/cli/tx.go
```diff
@@ -98,8 +98,14 @@ before running the command (e.g., via babylond create-bls-key).`))
...
 
-		txf := tx.NewFactoryCLI(clientCtx, cmd.Flags()).
-			WithTxConfig(clientCtx.TxConfig).WithAccountRetriever(clientCtx.AccountRetriever)
+		txf, err := tx.NewFactoryCLI(clientCtx, cmd.Flags())
+
+		if err != nil {
+			return err
+		}
+
+		txf = txf.WithTxConfig(clientCtx.TxConfig).WithAccountRetriever(clientCtx.AccountRetriever)
```
SebastianElvis, line 107: same ques here

SebastianElvis (approved):
Lgtm! CI fails though

x/zoneconcierge/keeper/proof_tx_in_block_test.go
KonradStaniec: not sure if I understand correctly, but it seems we need this assignemt as `txf` variable is used in subsequent operations like `txf.Prepare(val.ClientCtx)`, `txf.BuildUnsignedTx(msg)` etc.

x/checkpointing/client/cli/tx.go
KonradStaniec: here also we are using `txf` in subsequent operations.

vitsalis (approved):
Nice!

PR #360, opened 2023-04-24
Wire up grandpa decoder to babylon custom logic

KonradStaniec:
Hooks up picasso header decoding to babylon custom checkpointing logic

### DIFF
--- a/x/zoneconcierge/extended-client-keeper/keeper.go
+++ b/x/zoneconcierge/extended-client-keeper/keeper.go
@@ -1,8 +1,11 @@
 package extended_client_keeper
 
 import (
+	"bytes"
+
 	sdkerrors "cosmossdk.io/errors"
 	metrics "github.com/armon/go-metrics"
+	grandpa "github.com/babylonchain/babylon/x/zoneconcierge/types/grandpa"
 	"github.com/cometbft/cometbft/crypto/tmhash"
 	"github.com/cosmos/cosmos-sdk/codec"
 	storetypes "github.com/cosmos/cosmos-sdk/store/types"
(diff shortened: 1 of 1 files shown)


x/zoneconcierge/extended-client-keeper/keeper.go
```diff
@@ -35,6 +39,25 @@ func GetHeaderInfo(ctx sdk.Context, m exported.ClientMessage) *HeaderInfo {
 			ChaindId: msg.Header.ChainID,
 			Height:   uint64(msg.Header.Height),
 		}
+	case *ibcwasm.Header:
+		// try to decode wasm.data as a picasso header
+		gh, err := grandpa.DecodeHeader(grandpa.NewDecoder(bytes.NewReader(msg.Data)))
```
KonradStaniec, line 44: @blasrodri just to be sure, does `ibcwasm.Header.Data` field contains only serialized picasso header or there is some other data here ?

vmarkushin: It only contains [grandpa light-client header](https://github.com/ComposableFi/centauri/blob/ff3a592a108074b7f65956cbf6451d55abe147bd/light-clients/ics10-grandpa/src/client_message.rs#L42)

x/zoneconcierge/extended-client-keeper/keeper.go
```diff
@@ -35,6 +39,25 @@ func GetHeaderInfo(ctx sdk.Context, m exported.ClientMessage) *HeaderInfo {
...
+
+		if err == nil {
+			// this is header is unknown to babylon so we can't checkpoint it. Returning nil
+			// means that this light client will be frozen.
+			return nil
+		}
+
+		return &HeaderInfo{
```
blasrodri, line 52: don't you need any extra validation? Just asking, I'm totally new to it :)

x/zoneconcierge/extended-client-keeper/keeper.go
KonradStaniec: oh and how this header relates to the header parser that Eduardo implemented here - https://github.com/babylonchain/babylon/blob/centauri-integration-ibc-v7-stable/x/zoneconcierge/types/grandpa/header.go ?

x/zoneconcierge/extended-client-keeper/keeper.go
KonradStaniec: it does not need more validation as it is called only in our modified `UpdateClient` function after client message is already verified with `VerifyClientMessage` and `CheckForMisbehaviour`. 

This function should probably be package private, but this just a detail :)

PR #362, opened 2023-04-25
feat: support multiple chain ids in zoneconcierge chains info api 

gusin13:
### Description

This PR is part of the optimizations being done to improve babylon-api performance. `ChainInfo` rpc query is being called multiple times by the explorer apis. I have replaced this method with `ChainsInfo` which will accept a list of chain ids and will output latest header info of the concerned cz's

### Testing
Have added relevant fuzz tests and cli query


test/e2e/configurer/chain/queries.go
```diff
@@ -209,17 +209,6 @@ func (n *NodeConfig) QueryCheckpointChains() (*[]string, error) {
 	return &chainsResponse.ChainIds, nil
 }
 
-func (n *NodeConfig) QueryCheckpointChainInfo(chainId string) (*zctypes.ChainInfo, error) {
```
vitsalis, line 212: Why is this method removed?

x/zoneconcierge/keeper/grpc_query.go
```diff
@@ -37,24 +38,28 @@ func (k Keeper) ChainList(c context.Context, req *types.QueryChainListRequest) (
 	return resp, nil
 }
 
-// ChainInfo returns the latest info of a chain with given ID
-func (k Keeper) ChainInfo(c context.Context, req *types.QueryChainInfoRequest) (*types.QueryChainInfoResponse, error) {
+// ChainsInfo returns the latest info of chains with given IDs
```
vitsalis, line 41:
```suggestion
// ChainsInfo returns the latest info for a list of chains with given IDs
```

x/zoneconcierge/keeper/grpc_query.go
```diff
@@ -37,24 +38,28 @@ func (k Keeper) ChainList(c context.Context, req *types.QueryChainListRequest) (
...
 
-	if len(req.ChainId) == 0 {
-		return nil, status.Error(codes.InvalidArgument, "chain ID cannot be empty")
+	if len(req.ChainIds) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "chain IDs cannot be empty")
 	}
 
 	ctx := sdk.UnwrapSDKContext(c)
+	var chainsInfo []*types.ChainInfo
+	for _, chainID := range req.ChainIds {
```
vitsalis, line 53: Maybe also check if a chain ID is an empty string.

x/zoneconcierge/keeper/grpc_query_test.go
```diff
@@ -65,29 +66,52 @@ func FuzzChainList(f *testing.F) {
...
+				numForkHeaders: numForkHeaders,
+			})
+		}
 
-		// check if the chain info of is recorded or not
-		resp, err := zcKeeper.ChainInfo(ctx, &zctypes.QueryChainInfoRequest{ChainId: czChain.ChainID})
+		var chainIDs []string
+		for _, data := range chainsInfo {
+			chainIDs = append(chainIDs, data.chainID)
+		}
```
vitsalis, line 104: Maybe you could build the `chainIDs` list in the above `for` loop.

vitsalis (commented):
Nice!

x/zoneconcierge/keeper/grpc_query.go
gusin13: sure, fixed.

x/zoneconcierge/keeper/grpc_query_test.go
gusin13: sure, fixed.

test/e2e/configurer/chain/queries.go
gusin13: I couldn't find its usage in any of the existing e2e tests so removed it. 

@KonradStaniec I am not sure how to run e2e tests properly locally. I was getting this `warning: no tests to run`
Can you guide how to setup and add a new test in the suite?

proto/babylon/zoneconcierge/v1/query.proto
```diff
@@ -22,9 +22,9 @@ service Query {
     option (google.api.http).get = "/babylon/zoneconcierge/v1/chains";
   }
   // ChainInfo queries the latest info of a chain in Babylon's view
-  rpc ChainInfo(QueryChainInfoRequest) returns (QueryChainInfoResponse) {
+  rpc ChainsInfo(QueryChainsInfoRequest) returns (QueryChainsInfoResponse) {
     option (google.api.http).get =
-        "/babylon/zoneconcierge/v1/chain_info/{chain_id}";
+        "/babylon/zoneconcierge/v1/chains_info";
```
SebastianElvis, line 27: Is it possible to take a list of strings via the URL? Tbh this api is helpful for manually checking the liveness of checkpoints

x/zoneconcierge/client/cli/query.go
```diff
@@ -24,5 +23,28 @@ func GetQueryCmd(queryRoute string) *cobra.Command {
...
 
+	cmd.AddCommand(CmdChainsInfo())
+	return cmd
+}
+
+func CmdChainsInfo() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "chains-info <chain-ids>",
+		Short: "retrieve the chain info of given chain ids",
+		Args:  cobra.RangeArgs(1, 5),
```
SebastianElvis, line 34: Why do we put a limit of 5 here? maybe it's not needed

x/zoneconcierge/client/cli/query.go
```diff
@@ -24,5 +23,28 @@ func GetQueryCmd(queryRoute string) *cobra.Command {
 		RunE:                       client.ValidateCmd,
 	}
 
+	cmd.AddCommand(CmdChainsInfo())
+	return cmd
+}
+
+func CmdChainsInfo() *cobra.Command {
```
SebastianElvis, line 30: Thanks for adding this command!

x/zoneconcierge/client/cli/query.go
```diff
@@ -24,5 +23,28 @@ func GetQueryCmd(queryRoute string) *cobra.Command {
...
+	cmd := &cobra.Command{
+		Use:   "chains-info <chain-ids>",
+		Short: "retrieve the chain info of given chain ids",
+		Args:  cobra.RangeArgs(1, 5),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx := client.GetClientContextFromCmd(cmd)
+			queryClient := types.NewQueryClient(clientCtx)
+			req := types.QueryChainsInfoRequest{ChainIds: args}
+			resp, err := queryClient.ChainsInfo(context.Background(), &req)
```
SebastianElvis, line 39: can we use 
```suggestion
			resp, err := queryClient.ChainsInfo(cmd.Context(), &req)
```
here?

x/zoneconcierge/keeper/grpc_query.go
SebastianElvis: Could we move the check after L49? so that as long as there is an empty chain ID then no query will be performed

SebastianElvis (approved):
This would be a good optimisation! Just some minor comments

test/e2e/configurer/chain/queries.go
KonradStaniec: so to run tests:
1. first build docker container on branch with `make build-docker`
2. then run `make test-e2e` 
that should be it.

To add new tests, proper place is `e2e_test.go` file. One thing to note is that tests there are stateful i.e one needs to be aware what what is current state after previous tests.

x/zoneconcierge/keeper/grpc_query.go
```diff
@@ -37,24 +38,32 @@ func (k Keeper) ChainList(c context.Context, req *types.QueryChainListRequest) (
 	return resp, nil
 }
 
-// ChainInfo returns the latest info of a chain with given ID
-func (k Keeper) ChainInfo(c context.Context, req *types.QueryChainInfoRequest) (*types.QueryChainInfoResponse, error) {
+// ChainsInfo returns the latest info for a list of chains with given IDs
+func (k Keeper) ChainsInfo(c context.Context, req *types.QueryChainsInfoRequest) (*types.QueryChainsInfoResponse, error) {
```
KonradStaniec, line 42: Usually requests which returns lists should be paginated.  Here it is a bit tricky as we provide list of chain-ids and return list of chain info responses. In perfect world we would need to put some kinda of limit on number of chain-ids requested and tie it somehow to pagination. The limit is needed to avoid someone querying node for 1M of chain-ids, and exhausting node memory when it will try to read 1M `ChainInfo`  objects into memory.

Maybe for not just put limit in place ? (50 or 100 chain ids ?)

x/zoneconcierge/keeper/grpc_query.go
```diff
@@ -37,24 +38,32 @@ func (k Keeper) ChainList(c context.Context, req *types.QueryChainListRequest) (
...
+	for _, chainID := range req.ChainIds {
+		if len(chainID) == 0 {
+			return nil, status.Error(codes.InvalidArgument, "chain ID cannot be empty")
+		}
 
-	// find the chain info of this epoch
-	chainInfo, err := k.GetChainInfo(ctx, req.ChainId)
-	if err != nil {
-		return nil, err
+		chainInfo, err := k.GetChainInfo(ctx, chainID)
```
KonradStaniec, line 58: we are missing handling of the case when caller puts duplicated chain-id in the request list, imo it makes no sense to return duplicated chain-info results.

There two ways on handling that:
1. either deduplicate chain-ids when recveing requrest
2. or return error in case of first duplicate happening.

I think I would go for `2` as then usually duplicates in such case are callers error, and it would be good to inform caller about the error which happened.

x/zoneconcierge/keeper/grpc_query.go
gusin13: can you explain where to put after L49, currently it is at L54 and no query is performed if chain ID is empty

```
	if req == nil {
		return nil, status.Error(codes.InvalidArgument, "invalid request")
	}

	if len(req.ChainIds) == 0 {
		return nil, status.Error(codes.InvalidArgument, "chain IDs cannot be empty")
	}

	ctx := sdk.UnwrapSDKContext(c)
	var chainsInfo []*types.ChainInfo
	for _, chainID := range req.ChainIds {
		if len(chainID) == 0 {
			return nil, status.Error(codes.InvalidArgument, "chain ID cannot be empty")
		}
		......
	}
```

gusin13: sure, fixed.

proto/babylon/zoneconcierge/v1/query.proto
gusin13: Sure, I think it's already being passed that way in current setup. So the GET call would translate to `/babylon/zoneconcierge/v1/chains_info?chain_ids=1&chain_ids=2`

gusin13: I thought about this again, I couldn't test with multiple chain ids so I am not very clear if protobuf repeated strings would translate to HTTP get query params or body.

SebastianElvis: Yeah I could not think of any straightforward way either, nor could I find any practice for such APIs. So probably let's stick to your current change

x/zoneconcierge/keeper/grpc_query.go
gusin13: sure, good point. I have hardcoded a max limit of 100 now.

x/zoneconcierge/keeper/grpc_query.go
gusin13: good catch, sure I have implemented 2 now.

x/zoneconcierge/client/cli/query.go
gusin13: sure, have removed the limit now.

test/e2e/configurer/chain/queries.go
gusin13: Thanks @KonradStaniec, I have added this test and assertion back in the suite now.

gusin13:
@KonradStaniec @vitsalis @SebastianElvis Thanks for the reviews guys, I have addressed above comments. Please feel free to review again :)

KonradStaniec (approved):
Looks good !

At some point, we need to figure out how to best paginate those list queries so that caller have nice hints how to get next existing chain-ids, but at this point having limits should be enough 👍

vitsalis (approved):
Nice!

PR #365, opened 2023-04-27
feat: support multiple chain ids in FinalizedChainsInfo

gusin13:
### Description
This PR is part of the optimizations being done to improve babylon-api performance. 

`FinalizedChainInfo` rpc query is called multiple times by the explorer apis. I have replaced this method with `FinalizedChainsInfo` which will accept a list of chain ids so we can fetch info of multiple cz's at once. 

### Testing
Have added relevant fuzz tests and cli query


SebastianElvis (approved):
Nice!

x/zoneconcierge/keeper/grpc_query.go
```diff
@@ -184,54 +185,85 @@ func (k Keeper) ListEpochHeaders(c context.Context, req *types.QueryListEpochHea
...
+			return nil, status.Error(codes.InvalidArgument, "chain ID cannot be empty")
+		}
+
+		// check for duplicates and return error on first duplicate found
+		if encountered[chainID] {
+			return nil, status.Errorf(codes.InvalidArgument, "duplicate chain ID %s", chainID)
+		} else {
+			encountered[chainID] = true
+		}
 	}
```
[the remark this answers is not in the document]
gusin13: sure, fixed.

test/e2e/configurer/chain/queries.go
```diff
@@ -70,10 +63,10 @@ func (n *NodeConfig) QueryGRPCGateway(path string, parameters ...string) ([]byte
 	return bz, nil
 }
 
-// QueryBalancer returns balances at the address.
+// QueryBalances QueryBalancer returns balances at the address.
```
vitsalis, line 66:
```suggestion
// QueryBalances returns balances at the address.
```

test/e2e/configurer/chain/queries.go
```diff
@@ -41,12 +38,8 @@ func (n *NodeConfig) QueryGRPCGateway(path string, parameters ...string) ([]byte
 			return false
 		}
 
-		if len(parameters) > 0 {
-			q := req.URL.Query()
-			for i := 0; i < len(parameters); i += 2 {
-				q.Add(parameters[i], parameters[i+1])
-			}
-			req.URL.RawQuery = q.Encode()
+		if queryParams != nil {
```
vitsalis, line 41: I suggest that in the case that we do not want to pass any parameters, we do not pass `nil`, but instead pass an empty `queryParams` object.

x/zoneconcierge/keeper/grpc_query.go
```diff
@@ -184,54 +180,76 @@ func (k Keeper) ListEpochHeaders(c context.Context, req *types.QueryListEpochHea
...
-	}
+	for _, chainID := range req.ChainIds {
+		data := &types.FinalizedChainInfo{ChainId: chainID}
+
+		// find the last finalised chain info and the earliest epoch that snapshots this chain info
+		finalizedEpoch, chainInfo, err := k.GetLastFinalizedChainInfo(ctx, chainID)
+		if err != nil {
+			// skip if the finalised chain info is not found
+			if errors.Is(err, types.ErrEpochChainInfoNotFound) {
+				k.Logger(ctx).Error("finalised chain info not found", "chain ID", chainID)
```
vitsalis, line 215: No need to produce an error log for an invalid request. Could lead to someone hogging the logs by sending invalid requests.

x/zoneconcierge/keeper/grpc_query.go
```diff
@@ -184,54 +180,76 @@ func (k Keeper) ListEpochHeaders(c context.Context, req *types.QueryListEpochHea
...
-		return nil, err
-	}
+	for _, chainID := range req.ChainIds {
+		data := &types.FinalizedChainInfo{ChainId: chainID}
+
+		// find the last finalised chain info and the earliest epoch that snapshots this chain info
+		finalizedEpoch, chainInfo, err := k.GetLastFinalizedChainInfo(ctx, chainID)
+		if err != nil {
+			// skip if the finalised chain info is not found
+			if errors.Is(err, types.ErrEpochChainInfoNotFound) {
```
vitsalis, line 214: I don't like checking for particular errors as this might lead to inconsistencies if something minor is changed. I suggest we add a method `FinalizedChainInfoExists` or something similar, to check whether we don't need to add anything for this chain.

vitsalis (commented):
Nice work! Some suggestions:

test/e2e/configurer/chain/queries.go
gusin13: sure fixed

x/zoneconcierge/keeper/grpc_query.go
gusin13: sure fixed

KonradStaniec (approved):

x/zoneconcierge/keeper/grpc_query.go
gusin13: sure fixed

x/zoneconcierge/keeper/grpc_query.go
```diff
@@ -184,54 +179,78 @@ func (k Keeper) ListEpochHeaders(c context.Context, req *types.QueryListEpochHea
...
+	resp := &types.QueryFinalizedChainsInfoResponse{FinalizedChainsInfo: []*types.FinalizedChainInfo{}}
 
-	if err != nil {
-		return nil, err
-	}
+	for _, chainID := range req.ChainIds {
+		data := &types.FinalizedChainInfo{ChainId: chainID}
 
-	resp.RawCheckpoint = rawCheckpoint.Ckpt
+		exists, err := k.HasFinalizedChainInfo(ctx, chainID)
```
vitsalis, line 209: The `HasFinalizedChainInfo` method retrieves the latest finalized epoch each time. This is a storage read operation and is quite costly, especially given that this value won't change in-between iterations. I suggest we retrieve this value outside of this loop and instead of using the `HasFinalizedChainInfo(ctx, chainID)` function which is non-performant, we use a function  `EpochChainInfoExists(ctx, epochNum, chainID)`, which is similar to [this method](https://github.com/babylonchain/babylon/blob/dev/x/zoneconcierge/keeper/epoch_chain_info_indexer.go#L11) but only looks for existence.

x/zoneconcierge/keeper/epoch_chain_info_indexer.go
```diff
@@ -20,6 +21,23 @@ func (k Keeper) GetEpochChainInfo(ctx sdk.Context, chainID string, epochNumber u
 	return &chainInfo, nil
 }
 
+// HasFinalizedChainInfo checks if chain info exists for a given chain ID in the last finalised epoch
+func (k Keeper) HasFinalizedChainInfo(ctx sdk.Context, chainID string) (bool, error) {
```
vitsalis, line 25: This method is not that performant and does more than it should. I suggest we remove this and use something similar to [this method](https://github.com/babylonchain/babylon/blob/dev/x/zoneconcierge/keeper/epoch_chain_info_indexer.go#L11) to check for existence given a chain ID and epoch.

x/zoneconcierge/keeper/grpc_query.go
```diff
@@ -46,26 +48,19 @@ func (k Keeper) ChainsInfo(c context.Context, req *types.QueryChainsInfoRequest)
...
-
-		// check for duplicates and return error on first duplicate found
-		if encountered[chainID] {
-			return nil, status.Errorf(codes.InvalidArgument, "duplicate chain ID %s", chainID)
-		} else {
-			encountered[chainID] = true
-		}
+	// return if chain IDs contain duplicates or empty strings
+	if err := bbntypes.CheckForDuplicatesAndEmptyStrings(req.ChainIds); err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
```
vitsalis, line 63: The errors from the above function are not specific and helpful for the user. For example, "Duplicate string" might not lead the user to understand that they should not provide duplicate chain IDs.

x/zoneconcierge/keeper/grpc_query.go
gusin13: I am not sure how we will only check for existence as we have to perform [this](https://github.com/babylonchain/babylon/pull/365/files#diff-333a1b32e204c5ea9ac3117c0267aed6e36cd0399872d2ef28b4bc6ef756416fR70) check inside that function which will involve a storage read.

x/zoneconcierge/keeper/epoch_chain_info_indexer.go
gusin13: I am not sure how we will only check for existence as we have to perform [this](https://github.com/babylonchain/babylon/pull/365/files#diff-333a1b32e204c5ea9ac3117c0267aed6e36cd0399872d2ef28b4bc6ef756416fR70) check inside that function which will involve a storage read.

x/zoneconcierge/keeper/grpc_query.go
gusin13: sure, as `CheckForDuplicatesAndEmptyStrings` is a generic util and can be used in any module I have wrapped the error from the caller side instead. Have fixed it now..

x/zoneconcierge/keeper/epoch_chain_info_indexer.go
```diff
@@ -20,15 +21,16 @@ func (k Keeper) GetEpochChainInfo(ctx sdk.Context, chainID string, epochNumber u
...
 // GetLastFinalizedChainInfo gets the last finalised chain info recorded for a given chain ID
 // and the earliest epoch that snapshots this chain info
-func (k Keeper) GetLastFinalizedChainInfo(ctx sdk.Context, chainID string) (uint64, *types.ChainInfo, error) {
-	// find the last finalised epoch
-	finalizedEpoch, err := k.GetFinalizedEpoch(ctx)
-	if err != nil {
-		return 0, nil, err
-	}
-
+func (k Keeper) GetLastFinalizedChainInfo(ctx sdk.Context, chainID string, finalizedEpoch uint64) (uint64, *types.ChainInfo, error) {
```
vitsalis, line 33: Few things:
1. Why do we return a `uint64` corresponding to the finalized epoch if this is available from `chainInfo.LatestHeader.BabylonEpoch`?
2. This method seems exactly the same as `GetEpochChainInfo`, with the main difference that if the epoch parameter is different than the `chainInfo.LatestHeader.BabylonEpoch`, then we call it again with the epoch parameter set to `chainInfo.LatestHeader.BabylonEpoch`. So two questions:
      a. Why do we need to call `GetEpochChainInfo` again with the correct epoch? Does anything actually change on the `chainInfo` object?
      b. Maybe we can have a single method. This particular function does not need to be aware that it is retrieving something `finalized` as it just retrieves the data given to it in the epoch parameter. I would argue we don't actually need this method, especially if (a) doesn't change anything. If we do indeed need it, let's remove any reference to `finalized` as this function does not do anything finalized specific.

x/zoneconcierge/keeper/grpc_query.go
```diff
@@ -184,54 +179,80 @@ func (k Keeper) ListEpochHeaders(c context.Context, req *types.QueryListEpochHea
...
+	for _, chainID := range req.ChainIds {
+		data := &types.FinalizedChainInfo{ChainId: chainID}
 
-	// find the raw checkpoint and the best submission key for the finalised epoch
-	_, resp.BtcSubmissionKey, err = k.btccKeeper.GetBestSubmission(ctx, finalizedEpoch)
-	if err != nil {
-		return nil, err
-	}
+		// if the chain info is not found in the last finalised epoch, return the chain info with empty fields
+		if exists := k.EpochChainInfoExists(ctx, chainID, lastFinalizedEpoch); !exists {
```
vitsalis, line 216:
```suggestion
		if !k.EpochChainInfoExists(ctx, chainID, lastFinalizedEpoch) {
```

x/zoneconcierge/keeper/epoch_chain_info_indexer.go
```diff
@@ -20,15 +21,16 @@ func (k Keeper) GetEpochChainInfo(ctx sdk.Context, chainID string, epochNumber u
 	return &chainInfo, nil
 }
 
+// EpochChainInfoExists checks if the latest chain info exists of a given epoch for a given chain ID
+func (k Keeper) EpochChainInfoExists(ctx sdk.Context, chainID string, epochNumber uint64) bool {
```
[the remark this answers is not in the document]
gusin13: sure fixed

vitsalis (approved):

x/zoneconcierge/keeper/epoch_chain_info_indexer.go
gusin13: sure, fixed in my last commits.

PR #367, opened 2023-04-28
Fix bug in submission btc info

KonradStaniec:
While looking for ways of optimizing `BtcCheckpointInfo` endpoint I encountered bug in getting submission btc info, which could lead to misindexing of transaction.

This pr fixes the bug and provides set of tests triggering given problem.

In general I think we need ADR for describing how we decide which submission is the best. It can be done when we will start working on rewards.

### DIFF
--- a/x/btccheckpoint/keeper/keeper.go
+++ b/x/btccheckpoint/keeper/keeper.go
@@ -142,34 +142,57 @@ func (k Keeper) checkSubmissionStatus(ctx sdk.Context, info *types.SubmissionBtc
 
 func (k Keeper) GetSubmissionBtcInfo(ctx sdk.Context, sk types.SubmissionKey) (*types.SubmissionBtcInfo, error) {
 
-	var lowest uint64 = math.MaxUint64
-	var highest uint64 = uint64(0)
-	var lowestIndestInMostFreshBlock uint32 = math.MaxUint32
+	var youngestBlockDepth uint64 = math.MaxUint64
+
+	var lowestIndexInMostFreshBlock uint32 = math.MaxUint32
+
+	var oldestBlockDepth uint64 = uint64(0)
 
 	for _, tk := range sk.Key {
-		d, err := k.headerDepth(ctx, tk.Hash)
+		currentBlockDepth, err := k.headerDepth(ctx, tk.Hash)
 
 		if err != nil {
 			return nil, err
 		}
 
-		if d <= lowest {
-			lowest = d
-			if tk.Index < lowestIndestInMostFreshBlock {
-				lowestIndestInMostFreshBlock = tk.Index
-			}
+		if currentBlockDepth < youngestBlockDepth {
+			youngestBlockDepth = currentBlockDepth
+			lowestIndexInMostFreshBlock = tk.Index
 		}
 
-		if d > highest {
-			highest = d
+		// This case happens when we have two submissions in the same block.
+		if currentBlockDepth == youngestBlockDepth && tk.Index < lowestIndexInMostFreshBlock {
+			// This is something which needs a bit more careful thinking as it is used
+			// to determine which submission is better.
+			// Currently if two submissions of one checkpoint are in the same block,
+			// we pick tx with lower index as the point at which checkpoint happened.
+			// This is in line with the logic that if two submission are in the same block,
+			// they are esentially happening at the same time, so it does not really matter
+			// which index pick, and for possibble tie breaks it is better to pick lower one.
+			// This means in case when we have:
+			// Checkpoint submission `x` for epoch 5, both tx in same block at height 100, with indexes 1 and 10
+			// and
+			// Checkpoint submission `y` for epoch 5, both tx in same block at height 100, with indexes 3 and 9
+			// we will chose submission `x` as the better one.
+			// This good enough solution, but it is not perfect and leads to some edge cases like:
+			// Checkpoint submission `x` for epoch 5, one tx in block 99 with index 1, and second tx in block 100 with index 4
+			// and
+			// Checkpoint submission `y` for epoch 5, both tx in same block at height 100, with indexes 3 and 9
+			// In this case submission `y` will be better as it `earliest` tx in most fresh block is first. But at first glance
+			// submission `x` seems better.
+			lowestIndexInMostFreshBlock = tk.Index
+		}
+
+		if currentBlockDepth > oldestBlockDepth {
+			oldestBlockDepth = currentBlockDepth
 		}
 	}
 
 	return &types.SubmissionBtcInfo{
 		SubmissionKey:      sk,
-		OldestBlockDepth:   highest,
-		YoungestBlockDepth: lowest,
-		LatestTxIndex:      lowestIndestInMostFreshBlock,
+		OldestBlockDepth:   oldestBlockDepth,
+		YoungestBlockDepth: youngestBlockDepth,
+		LatestTxIndex:      lowestIndexInMostFreshBlock,
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


x/btccheckpoint/keeper/keeper.go
```diff
@@ -142,34 +142,57 @@ func (k Keeper) checkSubmissionStatus(ctx sdk.Context, info *types.SubmissionBtc
...
+			// Checkpoint submission `x` for epoch 5, both tx in same block at height 100, with indexes 1 and 10
+			// and
+			// Checkpoint submission `y` for epoch 5, both tx in same block at height 100, with indexes 3 and 9
+			// we will chose submission `x` as the better one.
+			// This good enough solution, but it is not perfect and leads to some edge cases like:
+			// Checkpoint submission `x` for epoch 5, one tx in block 99 with index 1, and second tx in block 100 with index 4
+			// and
+			// Checkpoint submission `y` for epoch 5, both tx in same block at height 100, with indexes 3 and 9
+			// In this case submission `y` will be better as it `earliest` tx in most fresh block is first. But at first glance
+			// submission `x` seems better.
```
vitsalis, line 182: How about we collect sets of `(firstPartBlockHeight, secondPartBlockHeight, firstPartTxIndex, secondPartTxIndex)` and then sort them based on an algorithm we define to choose the values that refer to the best submission? Will allow us to cleanly separate the selection algorithm and make it reusable by other people that are looking into the BTC chain to select the best Babylon chain. The second is especially very powerful as we want people to be able to efficiently differentiate the proper Babylon chain.

I think that in this case (even though those are the same checkpoints), we should clearly choose `y`. However, as long as everyone uses the same selection rule as us, we're good.

KonradStaniec: So there is already struct which is used for that purpose with `IsBetterThan` defined - https://github.com/babylonchain/babylon/blob/dev/x/btccheckpoint/types/types.go#L138. It does not contain all the data you mentioned as not all data is used currently for scoring purposes.

Current algorithm is pretty simple but definitely not the best i.e submission `A` is considered  better than submission `B` , if its most recent header is deeper in btc chain that submission `B` most recent header. In case of a tie (most recent header is the same one), submission with lowest txidx in most recent header wins.

In general I can find both advantages and disadvantages of this approach. (and several possible alternatives)

Imo current route for this problem is:
1. merge this pr as it fixes pretty insidious bug
2. create ADR/RFC spelling out possible alternatives. Also it would be nice to connect this problem to rewards because as soon you introduce rewards and relaying on transaction ordering to get those rewards, you must consider MEV i.e miners observing transactions in mempool and reordering them/injecting their own transactions to get those rewards.
3. After this ADR we can intrduce necessary data to `SubmissionBtcInfo` and tweak `IsBetterThan` implementation.

vitsalis: Makes sense, thanks for the explanation!

vitsalis (approved):

x/btccheckpoint/keeper/keeper_test.go
```diff
@@ -0,0 +1,64 @@
...
+		expectedOldestBlockDepth   uint64
+	}{
+		{"First header older. TxIndex larger in older header.", args{TxKeyDesc{TxIdx: 5, Depth: 10}, TxKeyDesc{TxIdx: 1, Depth: 0}}, 0, 1, 10},
+		{"First header older. TxIndex larger in younger header.", args{TxKeyDesc{TxIdx: 1, Depth: 10}, TxKeyDesc{TxIdx: 5, Depth: 0}}, 0, 5, 10},
+		{"Second header older. TxIndex larger in older header.", args{TxKeyDesc{TxIdx: 1, Depth: 0}, TxKeyDesc{TxIdx: 5, Depth: 10}}, 0, 1, 10},
+		{"Second header older. TxIndex larger in younger header.", args{TxKeyDesc{TxIdx: 5, Depth: 0}, TxKeyDesc{TxIdx: 1, Depth: 10}}, 0, 5, 10},
+		{"Same block. TxIndex larger in first transaction key.", args{TxKeyDesc{TxIdx: 5, Depth: 10}, TxKeyDesc{TxIdx: 1, Depth: 10}}, 10, 1, 10},
+		{"Same block. TxIndex larger in second transaction key.", args{TxKeyDesc{TxIdx: 1, Depth: 10}, TxKeyDesc{TxIdx: 5, Depth: 10}}, 10, 1, 10},
+	}
+	for _, tt := range tests {
```
vitsalis, line 38: Can we make this a fuzz test? Given that it depends on randomness.

KonradStaniec: tbh the only randomness here is for hashes for which we do not care that much. And I would prefer to have this particular test cases always checked as they trigger the bug. 

On the other hand fuzz test here is pretty good idea to check different combinations.

tldr; I have added separate fuzzing test for this method 😅

x/btccheckpoint/keeper/keeper_test.go
```diff
@@ -0,0 +1,120 @@
...
+			require.Equal(t, info.LatestTxIndex, tt.expectedTxIndex, tt.name)
+			require.Equal(t, info.OldestBlockDepth, tt.expectedOldestBlockDepth, tt.name)
+		})
+	}
+}
+
+func FuzzGetSubmissionBtcInfo(f *testing.F) {
+	f.Add(int64(1), uint32(0), uint32(1), uint32(1), uint32(1))
+
+	f.Fuzz(func(t *testing.T, seed int64, depth1 uint32, txidx1 uint32, depth2 uint32, txidx2 uint32) {
```
vitsalis, line 69: The style that we typically follow is we generate all the parameters from the randomness seed. e.g. we could have the following

```
depth1 := r.Uint32()
depth2 := r.Uint32()
txidx1 := r.Uint32()
txidx2 := r.Uint32()
for txidx2 == txidx1 {
    txidx2 = r.Uint32()
}
```

x/btccheckpoint/keeper/keeper_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+			require.Equal(t, info.YoungestBlockDepth, tt.expectedYoungestBlockDepth, tt.name)
+			require.Equal(t, info.LatestTxIndex, tt.expectedTxIndex, tt.name)
+			require.Equal(t, info.OldestBlockDepth, tt.expectedOldestBlockDepth, tt.name)
+		})
+	}
+}
+
+func FuzzGetSubmissionBtcInfo(f *testing.F) {
+	f.Add(int64(1), uint32(0), uint32(1), uint32(1), uint32(1))
```
vitsalis, line 67: If we had only the `seed` parameter here, we could dynamically generate `n` test cases each time with `datagen.AddRandomSeedsToFuzzer(f, n)`

vitsalis (commented):
Fuzz test looks good, although it does not follow our usual style. I guess this is related to our discussion about letting the Go fuzzer try to cover all paths.

x/btccheckpoint/keeper/keeper_test.go
KonradStaniec: hmm I think I missed this parts of discussion. How the style you describe is better than just go fuzzer generating this values for us ?

vitsalis: The reasoning is that we generate ourselves all of the data that we need for the tests through datagen methods and therefore we only need a seed. This allows us to only generate valid data (for example in this case different tx idxs) without having to skip due to the Go fuzzer providing invalid data. In this particular case this usefulness is not that obvious as we will rarely have to skip (given to how easy it is to generate different tx idxs), but for more complex data types (e.g. a BTC hash) we would skip most of the tests if we didn't generate proper data ourselves.

vitsalis: No problem with the way the test is structured right now, but I would prefer if it followed the style we have in the rest of the codebase.

KonradStaniec: Seems reasonable. I will adapt the test so it will not stand out  👍

PR #369, opened 2023-04-30
feat: Add support for multiple chain ids in EpochChainsInfo API

gusin13:
### Description

This PR is part of optimizations done to improve babylon-api performance. 

Explorer currently calls `EpochChainInfo` ( `/babylon/zoneconcierge/v1/chain_info/{chain_id}/epochs/{epoch_num}` ) multiple times for each chain. This PR replaces `EpochChainInfo` with `EpochChainsInfo` which can ingest a range of chain ids for a given epoch and retrieve epoch info for multiple chains at once. 

### Testing

Added fuzz , e2e tests and cli query.


test/e2e/e2e_test.go
```diff
@@ -30,28 +30,35 @@ func (s *IntegrationTestSuite) TestIbcCheckpointing() {
...
+
+	endEpoch, err := nonValidatorNode.QueryCheckpointForEpoch(endEpochNum)
 	s.NoError(err)
+	s.Equal(endEpoch.Status, ct.Finalized)
 
-	if epoch3.Status != ct.Finalized {
-		s.FailNow("Epoch 2 should be finalized")
-	}
+	// Check we have epoch info for opposing chain and some basic assertions
+	epochChainsInfo, err := nonValidatorNode.QueryEpochChainsInfo(3, []string{initialization.ChainBID})
```
vitsalis, line 50:
```suggestion
	epochChainsInfo, err := nonValidatorNode.QueryEpochChainsInfo(endEpoch, []string{initialization.ChainBID})
```

x/zoneconcierge/client/cli/query.go
```diff
@@ -24,13 +24,14 @@ func GetQueryCmd(queryRoute string) *cobra.Command {
...
 	cmd.AddCommand(CmdFinalizedChainsInfo())
+	cmd.AddCommand(CmdEpochChainsInfoInfo())
 	return cmd
 }
 
 func CmdChainsInfo() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "chains-info <chain-ids>",
-		Short: "retrieves the latest info for a list of chains with given IDs",
+		Short: "returns the latest info for a given list of chains",
```
vitsalis, line 34: Here we changed "retrieves" to "returns", but the method below still has "retrieves". tbh I prefer  the "retrieves" but np with both, but let's be consistent

x/zoneconcierge/client/cli/query.go
```diff
@@ -70,3 +71,31 @@ func CmdFinalizedChainsInfo() *cobra.Command {
...
+		Args:  cobra.ArbitraryArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx := client.GetClientContextFromCmd(cmd)
+			queryClient := types.NewQueryClient(clientCtx)
+
+			var epoch uint64
+			_, err := fmt.Sscanf(args[0], "%d", &epoch)
+			if err != nil {
+				return err
+			}
```
vitsalis, line 88:
```suggestion
			epoch, err := strconv.ParseUint(args[0], 10, 64)
			if err != nil {
				return err
			}
```

x/zoneconcierge/client/cli/query.go
```diff
@@ -70,3 +71,31 @@ func CmdFinalizedChainsInfo() *cobra.Command {
...
+
+func CmdEpochChainsInfoInfo() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "epoch-chains-info <epoch-num> <chain-ids>",
+		Short: "returns the latest info for a list of chains in a given epoch",
+		Args:  cobra.ArbitraryArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx := client.GetClientContextFromCmd(cmd)
+			queryClient := types.NewQueryClient(clientCtx)
+
```
vitsalis, line 83: We should check the number of arguments here. There should be at least 1 to specify the epoch number.

x/zoneconcierge/keeper/grpc_query_test.go
```diff
@@ -151,49 +152,53 @@ func FuzzHeader(f *testing.F) {
...
+			// prepare next epoch num and height
+			epochNum += datagen.RandomInt(r, 10) + 1
+			epochStartHeight += numHeaders
 		}
 
 		// attest the correctness of epoch info for each tested epoch
-		for i := uint64(0); i < numReqs; i++ {
-			resp, err := zcKeeper.EpochChainInfo(ctx, &zctypes.QueryEpochChainInfoRequest{EpochNum: epochNumList[i], ChainId: czChain.ChainID})
+		for epoch, info := range epochToChainInfo {
+			resp, err := zcKeeper.EpochChainsInfo(ctx, &zctypes.QueryEpochChainsInfoRequest{EpochNum: epoch, ChainIds: []string{info.chainID}})
```
vitsalis, line 197: Now that we have the capability to query for multiple chains we should take advantage of it. This test looks like the convertion of the 1-chainID case to work with the new syntax. Let's test the fact that we can query many different chains at the same time. Let's also test that we handle duplicates properly and reject stuff that's over the request limit.

vitsalis (approved):
Looks good! Some suggestions:

test/e2e/e2e_test.go
gusin13: sure fixed.

x/zoneconcierge/client/cli/query.go
gusin13: sure fixed.

x/zoneconcierge/keeper/grpc_query_test.go
gusin13: sure, have added more tests now.

vitsalis (approved):

PR #371, opened 2023-05-02
chore: circleci: Use image with go 1.20.3 installed

vitsalis:
On the [rpc-client repository](https://github.com/babylonchain/rpc-client/pull/33) I encountered issues due to the Ubuntu image being used having Go 1.18 installed instead of the 1.20 version that we use. These were fixed when I used an image with the appropriate version installed, so figured to do the same here.

Given that we use Docker for some of the build steps this fails. Will investigate further.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,26 +4,30 @@ version: 2.1
 
 orbs:
   aws-ecr: circleci/aws-ecr@8.2.1
+  go: circleci/go@1.7.3
+
 
 jobs:
   build-test:
-    docker:
-      - image: cimg/go:1.20.3
+    machine:
+      image: ubuntu-2204:2022.10.1
+      resource_class: large
     steps:
+      - go/install:
+          version: 1.20
       - checkout
       - run:
           name: Print Go environment
           command: "go env"
-      - restore_cache: # restores saved cache if no changes are detected since last run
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -5,15 +5,10 @@ version: 2.1
...
-# Define a job to be invoked later in a workflow.
-# See: https://circleci.com/docs/2.0/configuration-reference/#jobs
 jobs:
   build-test:
-    # Specify the execution environment. You can specify an image from Dockerhub or use one of our Convenience Images from CircleCI's Developer Hub.
-    # See: https://circleci.com/docs/2.0/configuration-reference/#docker-machine-macos-windows-executor
-    machine:
-      image: ubuntu-2204:2022.10.1
-      resource_class: large
+    docker:
```
KonradStaniec, line 10: Unfortunately for integration tests we are using and starting docker container, and last time I check it was impossible to do it in docker execution enviroment. That why we were always using https://circleci.com/docs/executor-intro/#linux-vm on babylon node side

vitsalis: I changed this to install the correct go version in the docker image and it looks like we encounter no issues 🥳

vitsalis:
Ok managed to resolve this using an orb, please check again!

PR #375, opened 2023-05-04
Add fuzz test for epoch finalization/confirmation

KonradStaniec:
Adds proper fuzz test for the logic which handles epoch finalization/confirmation.

### DIFF
--- a/x/btccheckpoint/keeper/msg_server_test.go
+++ b/x/btccheckpoint/keeper/msg_server_test.go
@@ -3,9 +3,12 @@ package keeper_test
 import (
 	"bytes"
 	"context"
+	"math"
 	"math/rand"
 	"testing"
 	"time"
+
+	"github.com/babylonchain/babylon/testutil/datagen"
 
 	dg "github.com/babylonchain/babylon/testutil/datagen"
 	keepertest "github.com/babylonchain/babylon/testutil/keeper"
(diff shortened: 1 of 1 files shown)


x/btccheckpoint/keeper/msg_server_test.go
```diff
@@ -589,3 +592,107 @@ func TestStateTransitionOfValidSubmission(t *testing.T) {
 		t.Errorf("Epoch Data missing of in unexpected state")
 	}
 }
+
+func FuzzConfirmAndDinalizeManyEpochs(f *testing.F) {
+	datagen.AddRandomSeedsToFuzzer(f, 20)
```
vitsalis, line 597:
```suggestion
	datagen.AddRandomSeedsToFuzzer(f, 10)
```
To follow similar tactic with other tests.

x/btccheckpoint/keeper/msg_server_test.go
```diff
@@ -589,3 +592,107 @@ func TestStateTransitionOfValidSubmission(t *testing.T) {
...
+
+	f.Fuzz(func(t *testing.T, seed int64) {
+		r := rand.New(rand.NewSource(seed))
+		tk := InitTestKeepers(t)
+		defaultParams := btcctypes.DefaultParams()
+		kDeep := defaultParams.BtcConfirmationDepth
+		wDeep := defaultParams.CheckpointFinalizationTimeout
```
vitsalis, line 604:
```suggestion
		kDeep := r.Intn(10) + 1
		wDeep := kDeep + r.Intn(100) + 1
```

x/btccheckpoint/keeper/msg_server_test.go
```diff
@@ -589,3 +592,107 @@ func TestStateTransitionOfValidSubmission(t *testing.T) {
...
+				if epoch <= uint64(numFinalizedEpochs) {
+					tk.BTCLightClient.SetDepth(blck1.HeaderBytes.Hash(), int64(finalizationDepth))
+					finalizationDepth = finalizationDepth - 1
+					tk.BTCLightClient.SetDepth(blck2.HeaderBytes.Hash(), int64(finalizationDepth))
+
+					// first submission is always deepest one, and second block is the most recent one
+					if j == 1 {
+						bestSumbissionInfos[epoch] = uint64(finalizationDepth)
+					}
+					finalizationDepth = finalizationDepth - 1
```
vitsalis, line 640: We are executing this statement two times in this `if` condition, why?

x/btccheckpoint/keeper/msg_server_test.go
```diff
@@ -589,3 +592,107 @@ func TestStateTransitionOfValidSubmission(t *testing.T) {
...
+				} else if epoch <= uint64(numFinalizedEpochs+numConfirmedEpochs) {
+					tk.BTCLightClient.SetDepth(blck1.HeaderBytes.Hash(), int64(confirmationDepth))
+					confirmationDepth = confirmationDepth - 1
```
vitsalis, line 643: What if `confirmationDepth` (initially `wDeep - 1`) starts less than `numConfirmedEpochs  * 2` (i.e. the number of times this subtraction, along with the next one are going to be executed) ?

vitsalis (commented):
Slightly confused about the logic of this:

x/btccheckpoint/keeper/msg_server_test.go
KonradStaniec: tbh this was deliberate, as I consider this logic quite a bit important and the tests are pretty fast so there is no downside from going to 20 iterations

x/btccheckpoint/keeper/msg_server_test.go
KonradStaniec: because each state (Finalize, Confirmed, Submitted) have different variable to track depth, so to register proper depth for given epoch we need to know what will be to final state of the epoch.

x/btccheckpoint/keeper/msg_server_test.go
KonradStaniec: not sure I understand, but if `confirmationDepth` would start too low then most probably test would fail, as epoch which we expected to be confirmed would no be.

In general in this test there are three depth intervals:
form math.MaxUint32 to wDeep - those depth will be finalized
from  wDepth - 1 to kDeep - those depths will be confirmed
from kDepp - 1 to 0 - those depths will be submitted
And whole data generation is adjusted so that expected blocks, ends in correct intervals

x/btccheckpoint/keeper/msg_server_test.go
KonradStaniec: those parameters needs to be the same as the ones used in `btccheckpoint.Keeper` created by `InitTestKeepers(t)`. And `InitTestKeepers` uses default parameters.

vitsalis (approved):
Thanks for the explanations!

PR #381, opened 2023-05-16
Bump wasmd to rc2

KonradStaniec:
Small pr to bump wasmd to rc2 versions.

One thing I noticed, both ibc-go and wasmd uses `FeeKeeper` as their `ICS4Wrapper` implementation in their test-apps (https://github.com/cosmos/ibc-go/blob/main/testing/simapp/app.go#LL413C6-L413C18), where we only use `ChannelKeeper`. 

From what I understand `FeeKeeper` is actually enables https://github.com/cosmos/ibc/tree/main/spec/app/ics-029-fee-payment, which is scheme used to pay rewards to relayers. It seems like a good idea to enable it also on Babylon . What do you think @SebastianElvis  ?

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -635,6 +635,7 @@ func NewBabylonApp(
 		app.StakingKeeper,
 		distrkeeper.NewQuerier(app.DistrKeeper),
 		app.IBCKeeper.ChannelKeeper,
+		app.IBCKeeper.ChannelKeeper,
 		&app.IBCKeeper.PortKeeper,
 		scopedWasmKeeper,
 		app.TransferKeeper,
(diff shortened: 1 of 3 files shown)


SebastianElvis:
> From what I understand FeeKeeper is actually enables [cosmos/ibc@main/spec/app/ics-029-fee-payment](https://github.com/cosmos/ibc/tree/main/spec/app/ics-029-fee-payment?rgh-link-date=2023-05-16T06%3A28%3A44Z), which is scheme used to pay rewards to relayers. It seems like a good idea to enable it also on Babylon . What do you think @SebastianElvis ?

Agreed, we will need this to incentivise people to run relayers anyway

SebastianElvis (approved):

PR #387, opened 2023-05-24
zoneconcierge: adding header timestamp to IndexedHeader

SebastianElvis:
This PR adds the CZ header's timestamp to `IndexedHeader`. It's needed for phase 3 integration, namely notifying CZ to unbond mature validators before this timestamp.


proto/babylon/zoneconcierge/v1/zoneconcierge.proto
```diff
@@ -19,15 +21,19 @@ message IndexedHeader {
   // height is the height of this header on CZ ledger
   // (hash, height) jointly provides the position of the header on CZ ledger
   uint64 height = 3;
+  // time is the timestamp of this header on CZ ledger
+  // it is needed for CZ to unbond all mature validators/delegations
+  // before this timestamp when this header is BTC-finalised
+  google.protobuf.Timestamp time = 4 [ (gogoproto.stdtime) = true ];
```
vitsalis, line 27: In our own [modified staking](https://github.com/babylonchain/babylon/blob/dev/x/epoching/keeper/modified_staking.go#L29) we only use the block and we retrieve time from there. Why do we need to add the time here?

SebastianElvis: > In our own [modified staking](https://github.com/babylonchain/babylon/blob/dev/x/epoching/keeper/modified_staking.go?rgh-link-date=2023-05-24T07%3A17%3A19Z#L29) we only use the block and we retrieve time from there.

Yes this is because Babylon checkpoints itself to Bitcoin and has access to the epoch boundary header (including height and time).

> Why do we need to add the time here? Can you provide an example of how this is going to be used?

The situation is a bit different in the CZ side (I realised this during developing Babylon contract). The `ApplyMatureUnbonding` [function](https://github.com/babylonchain/wasmbinding/blob/main/utils/staking.go#L19) in CZ side needs to use both height and time, which can only come from BTC-finalised `IndexedHeader`. However, The `IndexedHeader` does not have time at the moment. The current Babylon contract uses [time of the Babylon header](https://github.com/babylonchain/babylon-contract/blob/debug-lifecycle/src/state/mod.rs#L56-L61) that timestamps this CZ header, which is not accurate.

vitsalis (approved):

PR #390, opened 2023-05-26
e2e: merge integration tests with e2e tests

SebastianElvis:
This PR proposes to merge integration tests with e2e tests. The goal of the two tests are largely the same. The `e2e` covers even more scenarios since 
- `e2e` invokes txs via CLI while integration tests do direct invocation.
- `e2e` tests phase1 integration.

Specifically, the integration tests do the following:

- `TestTestnetRuninng`: query interfaces to ensure the chain is running
- `TestBtcLightClientGenesis`: query the base header in BTC light client and verify correctness
- `TestNodeProgress`: query epoch number and ensure its correctness
- `TestSendTx`: send a BTC header and see if it can get into BTC light client
- `TestFailInvalidBTCTransactions`: ensure bogus BTC header and checkpoint txs will be rejected by Babylon
- `TestSubmitCheckpoint`: ensure correct checkpoint can be accepted by Babylon
- `TestConfirmCheckpoint`: ensure deep enough checkpoint becomes confirmed on Babylon

Here I migrated `TestBtcLightClientGenesis`  and `TestSendTx` to `e2e`. The other tests are already covered by `e2e`:

- `TestTestnetRuninng` is covered by all the tests that make queries and wait for blocks
- `TestNodeProgress` is covered by all scenarios that involve `FinalizeSealedEpochs`
- `TestSubmitCheckpoint` and `TestConfirmCheckpoint` are covered by assertions in ` TestIbcCheckpointing()`

Meanwhile, `TestFailInvalidBTCTransactions` is a bit overlapped with our fuzz tests so I did not migrate it.

Opinions or feedbacks would be helpful. If we are okay with this, then I will remove code for integration tests and remove the corresponding step in CI. cc @vitsalis @KonradStaniec


KonradStaniec:
to fix CI you need to add `build-docker` step to CI. Earlier it was built by `integration-tests`  step

.circleci/config.yml
```diff
@@ -32,12 +32,10 @@ jobs:
...
             make test
-      - run: # sudo is needed, so that integration test binary have proper access to nodes keyring
-          name: Run integration tests
+      - run:
+          name: Build docker image
           command: |
-            sudo -E env "PATH=$PATH" make localnet-test-integration
-      # TODO: If CI tests will take to long consider having only this e2e test
-      # instead of separate integration tests and e2e tests.
+            make build-docker
```
vitsalis, line 38: Given that e2e tests require a docker image to be available, maybe we can build the Docker image as part of the `make test-e2e` command? We did a similar thing with `make localnet-test-integration`.

vitsalis (approved):
Nice!

.circleci/config.yml
SebastianElvis: Good idea. `make test-e2e` also requires an updated docker image so it makes sense for development as well. Fixed

KonradStaniec (approved):

PR #393, opened 2023-05-30
hotfix: fixing marshaling of `MsgWrappedCreateValidator` and a flaky test

SebastianElvis:
This PR fixes two bugs:

- Marshaling of MsgWrappedCreateValidator: The Pubkey inside MsgWrappedCreateValidator.MsgCreateValidator is in type Any, thus requires extra handling for marshaling/unmarshaling. Specifically, MsgWrappedCreateValidator needs to implement UnpackInterfacesMessage interface. This PR also adds a test ensuring the decoding is correct. This PR is largely based on [#215 (files)](https://github.com/babylonchain/babylon/pull/215/files).
- Flaky test: FuzzFinalizedChainInfo is flaky because the chainIDs sometimes contains duplicated value. Worked around this by enforcing a lower bound on random chain ID length.


x/zoneconcierge/keeper/grpc_query_test.go
```diff
@@ -417,7 +417,7 @@ func FuzzFinalizedChainInfo(f *testing.F) {
 		)
 		numChains := datagen.RandomInt(r, 100) + 1
 		for i := uint64(0); i < numChains; i++ {
-			czChainIDLen := datagen.RandomInt(r, 50) + 1
+			czChainIDLen := datagen.RandomInt(r, 50) + 100
```
vitsalis, line 420: This does not completely solve the issue as you can still have the same chain ID albeit with a lower probability. I suggest we create a hashmap or something to check for existing values and recreate the chainID until we get a unique one.

vitsalis (approved):
Some concerns about the fix of the flaky test. Overall, I recommend that we don't include the fix to the flaky test here as it should not be part of the hotfix, it can be another PR.

PR #395, opened 2023-05-31
e2e: basic e2e tests for phase2 integration

SebastianElvis:
This PR provides a basic e2e test for phase2 integration. The test case deploys a Babylon contract on chain B and establishes an IBC channel between chain A and chain B's Babylon contract.

Along the way, this PR refactors the IBC functionalities in `e2e` a bit. Will need to add more test cases for phase2 in subsequence PRs.


SebastianElvis:
@KonradStaniec @vitsalis Guys I may need some suggestions here. The `e2e` step shows that all test cases passed, while the overall test suite shows the test failed. I have debugged for a long time and did not have any idea. Any thoughts? Thanks

test/e2e/e2e_test.go
```diff
@@ -100,42 +107,74 @@ func (s *IntegrationTestSuite) TestIbcCheckpointing() {
...
+	s.Len(openedChannels, 2)
+
+	// wait for 1 sealed epochs
+	currEpoch, err := nonValidatorNode.QueryCurrentEpoch()
+	s.NoError(err)
+	endEpoch := currEpoch + 1
+	nonValidatorNode.WaitUntilHeight(int64(initialization.BabylonEpochInterval*endEpoch + 3))
+
+	// Finalize the 1 sealed epochs
+	nonValidatorNode.FinalizeSealedEpochs(currEpoch, endEpoch)
```
KonradStaniec, line 154: problem is here: you need to start finalizing epochs from oldest non-finalized epoch i.e `4`, if you start finalizing from `currEpoch` you may as well start from epoch `10` so you will have gap from epoch `3` to `10`.

Then submitting checkpoint will succeed i.e it will enter mempool (no stateful checks) but it will fail during execution so this - https://github.com/babylonchain/babylon/blob/dev/test/e2e/configurer/chain/commands.go#L150 will time out. That is also reason why the tests take that long, for each checkpoint submission we wait until time hit. If you start finalizing from 4, the testing time drops by half.

I am not sure why `Errorf` called due to timeout, do not show that test failure though 🤷

KonradStaniec: changing `ErrorF` to `FatalF` produces this outcome:

```
ode.go:196: [2m54.956s] successfully inserted btc spv proofs. From container bbn-test-a-node-babylon-default-a-3
    node.go:136: node bbn-test-a-node-babylon-default-a-3 timed out waiting for condition. Msg: Checkpoint should be submitted
=== NAME  TestIntegrationTestSuite/TestPhase2_BabylonContract
    testing.go:1471: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
```

SebastianElvis:
Thanks @KonradStaniec for identifying the bug. Now this PR is ready for review.

One thing is that the IBC packets take quite long to be relayed and I haven't found a good way for testing BTC timestamps in IBC packets. Thus I narrowed down the scope of this PR a bit, to adding testing infra for phase2 and a basic test case for deploying Babylon contract and establishing IBC channel between chain A and Babylon contract. Will find a way for testing consecutive BTC timestamps in a subsequent PR.

test/e2e/configurer/base.go
```diff
@@ -163,26 +170,59 @@ func (bc *baseConfigurer) runIBCRelayer(chainConfigA *chain.Config, chainConfigB
...
+// It returns the contract address
+func (bc *baseConfigurer) DeployWasmContract(contractCodePath string, chain *chain.Config, initMsg string) (string, error) {
+	nonValidatorNode, err := chain.GetNodeAtIndex(2) // TODO: find a generic way to get non validator node
+	if err != nil {
+		return "", err
+	}
+
+	// store wasm code
+	nonValidatorNode.StoreWasmCode(contractCodePath, initialization.ValidatorWalletName)
+	nonValidatorNode.WaitForNextBlocks(3)
```
KonradStaniec, line 198: why 3 ? 2 or 1 is not enugh to avoid the sequence mismatch errors ?

test/e2e/configurer/base.go
```diff
@@ -163,26 +170,59 @@ func (bc *baseConfigurer) runIBCRelayer(chainConfigA *chain.Config, chainConfigB
...
+
+	// instantiate contract with the wasm code ID
+	latestWasmCodeID := int(nonValidatorNode.QueryLatestWasmCodeID())
+	if latestWasmCodeID == 0 {
```
KonradStaniec, line 202: hmm this is now used in two test meaning, in first one codeid will be 1 and in second it will be 2. It would be better to check what is latest codeid before code store and check here wheter it was incremented

test/e2e/configurer/chain/commands.go
```diff
@@ -101,6 +101,9 @@ func (n *NodeConfig) InsertProofs(p1 *btccheckpointtypes.BTCSpvProof, p2 *btcche
 	n.LogActionF("successfully inserted btc spv proofs")
 }
 
+// FinalizeSealedEpochs finalises epochs between [startEpoch, lastEpoch]
```
KonradStaniec, line 104: nitpick: we use american version everywhere i.e `finalizes`

SebastianElvis: Ok then 🙈 Need to change this in particular for epoching and zoneconcierge as well. Probably in another PR

test/e2e/configurer/base.go
SebastianElvis: Good point. Fixed

test/e2e/configurer/base.go
SebastianElvis: Nice catch, fixed

KonradStaniec (approved):

PR #399, opened 2023-07-14
Bump comet bft

KonradStaniec:
bumps comet bft  to 0.37.2 to fix - https://github.com/babylonchain/babylon/security/dependabot/16


vitsalis (approved):
Should this be a hotfix and deployed to the alpha testnet? Is it possible to apply it without a gov proposal?

KonradStaniec:
As discussed on our call, we do not need hot fix right now.

PR #402, opened 2023-12-16
README: updating discord link

GakiBash:
Hi, reading the doc I found out the current discord link is down, so I figured I would correct it myself instead of opening an issue and requiring someone to open a PR.

I found the new discord link in the Babylon's twitter bio.

That's it, doesn't require testing or changelog entry.


SebastianElvis (approved):
Thank you!

PR #408, opened 2024-01-14
Few typos fix

dzizazda:


x/btccheckpoint/types/btccheckpoint.pb.go
```diff
@@ -321,15 +321,15 @@ func (m *TransactionInfo) GetProof() []byte {
 }
 
 // TODO: Determine if we should keep any block number or depth info.
-// On one hand it may be usefull to determine if block is stable or not, on
```
SebastianElvis, line 324: This file is also generated from protobuf files. To address this we need to change the corresponding `btccheckpoint.proto` file and do `make proto-gen`

SebastianElvis:
Closed due to staleness

PR #418, opened 2024-01-22
chore: add parameter sections for module docs

SebastianElvis:


x/btcstaking/README.md
```diff
@@ -91,6 +90,48 @@ delegation unbonded immediately upon such a signature.
 
 The BTC Staking module maintains the following KV stores.
 
+### Parameters
+
+The [parameter storage](./keeper/params.go) maintains the BTC Staking module's
+parameters. The BTC Staking module's parameters are represented as a `Params`
+[object](../../proto/babylon/btcstaking/v1/params.proto) defined as follows:
```
vitsalis, line 97: Maybe we should not link to proto files (or put their names) here. It will make copy pasting the docs in our docs website not work.

vitsalis (approved):

x/btcstaking/README.md
SebastianElvis: I'd prefer to keep them in the docs here since these will give clear messages on where these objects are defined. For copying/pasting docs to the docs website, how about we create some bash scripts to this end, i.e., a bash script that copies each doc to the docs repo, and replaces `../../` with appropriate URLs. Wdyt?

vitsalis: Yes we could

PR #420, opened 2024-01-22
BTC time-stamping phase 2 tests follow-up

maurolacy:
A small PR with some improvements  / follow-up of #419.

This uses a pre-built public `cosmos-relayer` image to try and save some time during e2e tests. The image _has_ to be built for the same platform the CI is running on (`amd64`), or the phase 2 tests will fail with obscure error messages.

### DIFF
--- a/test/e2e/btc_timestamping_phase2_rly_test.go
+++ b/test/e2e/btc_timestamping_phase2_rly_test.go
@@ -105,8 +105,9 @@ func (s *BTCTimestampingPhase2RlyTestSuite) Test1IbcCheckpointingPhase2Rly() {
 	}, time.Minute, time.Second*2)
 
 	// ensure the next receive sequence number of Babylon contract is also 3
+	var nextSequenceRecv *channeltypes.QueryNextSequenceReceiveResponse
 	s.Eventually(func() bool {
-		nextSequenceRecv, err := czNode.QueryNextSequenceReceive(babylonChannel.Counterparty.ChannelId, babylonChannel.Counterparty.PortId)
+		nextSequenceRecv, err = czNode.QueryNextSequenceReceive(babylonChannel.Counterparty.ChannelId, babylonChannel.Counterparty.PortId)
 		if err != nil {
 			return false
 		}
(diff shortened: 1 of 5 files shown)


test/e2e/containers/config.go
```diff
@@ -15,7 +15,7 @@ const (
 	hermesRelayerRepository = "informalsystems/hermes"
 	// TODO: Replace with version tag once we have a working version
 	hermesRelayerTag        = "master"
-	cosmosRelayerRepository = "babylonchain/cosmos-relayer"
+	cosmosRelayerRepository = "public.ecr.aws/t9e9i3h0/cosmos-relayer"
```
SebastianElvis, line 18: Where did you find this image? Would be good if we have a link in a comment here

test/e2e/btc_timestamping_phase2_rly_test.go
```diff
@@ -105,8 +125,9 @@ func (s *BTCTimestampingPhase2RlyTestSuite) Test1IbcCheckpointingPhase2Rly() {
 	}, time.Minute, time.Second*2)
 
 	// ensure the next receive sequence number of Babylon contract is also 3
```
SebastianElvis, line 127:
```suggestion
	// ensure the next receive sequence number of Babylon contract is also 4
```

test/e2e/btc_timestamping_phase2_rly_test.go
```diff
@@ -105,8 +125,9 @@ func (s *BTCTimestampingPhase2RlyTestSuite) Test1IbcCheckpointingPhase2Rly() {
...
+	var nextSequenceRecv *channeltypes.QueryNextSequenceReceiveResponse
 	s.Eventually(func() bool {
```
SebastianElvis, line 129: At this point `nextSequenceRecv` is already 4 and the node is supposed to be accessible. Does `QueryNextSequenceReceive` need to be in `Eventual`?

test/e2e/containers/config.go
maurolacy: Basically, I built this using the `build-cosmos-relayer-docker` target on a amd64 machine, tagged it, and uploaded to that registry (which @filippos47 grant me access to). I will add a comment.

test/e2e/btc_timestamping_phase2_rly_test.go
maurolacy: It was that way originally. But yes, I can try to remove it.

SebastianElvis (approved):
Lgtm!

PR #430, opened 2024-01-26
chore: Remove leftover signer config

vitsalis:
This was used for the BLS signer process. Don't think it's being used anywhere else.

### DIFF
--- a/cmd/babylond/cmd/custom_babylon_config.go
+++ b/cmd/babylond/cmd/custom_babylon_config.go
@@ -7,12 +7,6 @@ import (
 	bbn "github.com/babylonchain/babylon/types"
 )
 
-const (
-	defaultKeyName       = ""
-	defaultGasPrice      = "0.01ubbn"
-	defaultGasAdjustment = 1.5
-)
-
 type BtcConfig struct {
 	Network string `mapstructure:"network"`
 }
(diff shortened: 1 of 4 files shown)


gitferry (commented):
Nice work! I was thinking maybe we can remove the `client.toml` entirely as both the checkpointing module and zone concierage module do not need it. Relevant code is in `babylon/app/utils.go`. And, maybe we can also remove the relevant part in [doc](https://docs.babylonchain.io/docs/user-guides/btc-timestamping-testnet/become-validator#3-modify-the-configuration)

vitsalis:
> I was thinking maybe we can remove the client.toml entirely in this PR as both the checkpointing module and zone concierge module do not need it. The relevant code is in babylon/app/utils.go.

From my understanding, `client.toml` is used by entities using the `babylond` executable as a client. For example, osmosis maintains a `client.toml` as well and I don't think we add any extra values there. Eitherway, if to remove it, I think it's better to do it in another PR. Is there anything else from `babylon/app/utils.go` that you would like to see removed relevant to this PR?

> And, maybe we can also remove the relevant part in [doc](https://docs.babylonchain.io/docs/user-guides/btc-timestamping-testnet/become-validator#3-modify-the-configuration)

This is a legacy doc that we are keeping for now as the old testnet is still running. It will be replaced by [this one](https://docs.babylonchain.io/docs/user-guides/btc-staking-testnet/become-validator)

gitferry (approved):
All good from my side

PR #431, opened 2024-01-26
Fix bip322 pop

KonradStaniec:
- add bip322 sig generation for p2tr address and p2wpkh  address
- improve types in api
- properly check that btc address used in bip322, corresponds to bip340 public key

TODO:
- [x] more tests
- [x] refactor a bit

### DIFF
--- a/crypto/bip322/bip322.go
+++ b/crypto/bip322/bip322.go
@@ -76,23 +76,10 @@ func toSignPkScript() []byte {
 	return script
 }
 
-// addressToPkScript takes an address and creates a payment script for it
-func addressToPkScript(addr string, net *chaincfg.Params) ([]byte, error) {
-	decoded, err := btcutil.DecodeAddress(addr, net)
-	if err != nil {
-		return nil, err
-	}
-	pkScript, err := txscript.PayToAddrScript(decoded)
-	if err != nil {
-		return nil, err
-	}
-	return pkScript, nil
-}
-
 // GetToSpendTx builds a toSpend transaction based on the BIP-322 spec
 // https://github.com/bitcoin/bips/blob/e643d247c8bc086745f3031cdee0899803edea2f/bip-0322.mediawiki#full
 // It requires as input the message that is signed and the address that produced the signature
-func GetToSpendTx(msg []byte, address string, net *chaincfg.Params) (*wire.MsgTx, error) {
+func GetToSpendTx(msg []byte, address btcutil.Address) (*wire.MsgTx, error) {
 	toSpend := wire.NewMsgTx(toSpendVersion)
 	toSpend.LockTime = toSpendLockTime
 
@@ -114,10 +101,12 @@ func GetToSpendTx(msg []byte, address string, net *chaincfg.Params) (*wire.MsgTx
 
 	// Create the output
 	// The PK Script should be a pay to addr script on the provided address
-	pkScript, err := addressToPkScript(address, net)
+	pkScript, err := txscript.PayToAddrScript(address)
+
 	if err != nil {
 		return nil, err
 	}
+
 	output := wire.NewTxOut(toSpendOutputValue, pkScript)
 
 	toSpend.AddTxIn(input)
@@ -128,7 +117,7 @@ func GetToSpendTx(msg []byte, address string, net *chaincfg.Params) (*wire.MsgTx
 // GetToSignTx builds a toSign transaction based on the BIP-322 spec
 // https://github.com/bitcoin/bips/blob/e643d247c8bc086745f3031cdee0899803edea2f/bip-0322.mediawiki#full
 // It requires as input the toSpend transaction that it spends and the message signature
-func GetToSignTx(toSpend *wire.MsgTx) (*wire.MsgTx, error) {
+func GetToSignTx(toSpend *wire.MsgTx) *wire.MsgTx {
 	toSign := wire.NewMsgTx(toSignVersion)
 	toSign.LockTime = toSignLockTime
 
@@ -145,26 +134,21 @@ func GetToSignTx(toSpend *wire.MsgTx) (*wire.MsgTx, error) {
 
 	toSign.AddTxIn(input)
 	toSign.AddTxOut(output)
-	return toSign, nil
+	return toSign
 }
 
-func Verify(msg []byte, sig []byte, address string, net *chaincfg.Params) error {
-	toSpend, err := GetToSpendTx(msg, address, net)
+func Verify(
+	msg []byte,
+	witness wire.TxWitness,
+	address btcutil.Address,
+	net *chaincfg.Params) error {
+
+	toSpend, err := GetToSpendTx(msg, address)
 	if err != nil {
 		return err
 	}
 
-	// Convert the signature into a witness stack
-	witness, err := simpleSigToWitness(sig)
-
-	if err != nil {
-		return err
-	}
-
-	toSign, err := GetToSignTx(toSpend)
-	if err != nil {
-		return err
-	}
+	toSign := GetToSignTx(toSpend)
 
 	toSign.TxIn[0].Witness = witness
 
@@ -225,17 +209,13 @@ func SignWithP2WPKHAddress(
 		return nil, err
 	}
 
-	toSpend, err := GetToSpendTx(msg, witnessAddr.EncodeAddress(), net)
+	toSpend, err := GetToSpendTx(msg, witnessAddr)
 
 	if err != nil {
 		return nil, err
 	}
 
-	toSign, err := GetToSignTx(toSpend)
-
-	if err != nil {
-		return nil, err
-	}
+	toSign := GetToSignTx(toSpend)
 
 	fetcher := txscript.NewCannedPrevOutputFetcher(
 		toSpend.TxOut[0].PkScript,
@@ -268,17 +248,13 @@ func SignWithP2TrSpendAddress(
 		return nil, err
 	}
 
-	toSpend, err := GetToSpendTx(msg, witnessAddr.EncodeAddress(), net)
+	toSpend, err := GetToSpendTx(msg, witnessAddr)
 
 	if err != nil {
 		return nil, err
 	}
 
-	toSign, err := GetToSignTx(toSpend)
-
-	if err != nil {
-		return nil, err
-	}
+	toSign := GetToSignTx(toSpend)
 
 	fetcher := txscript.NewCannedPrevOutputFetcher(
 		toSpend.TxOut[0].PkScript,
--- a/crypto/bip322/bip322.go
+++ b/crypto/bip322/bip322.go
@@ -116,7 +116,9 @@ func GetToSpendTx(msg []byte, address btcutil.Address) (*wire.MsgTx, error) {
 
 // GetToSignTx builds a toSign transaction based on the BIP-322 spec
 // https://github.com/bitcoin/bips/blob/e643d247c8bc086745f3031cdee0899803edea2f/bip-0322.mediawiki#full
-// It requires as input the toSpend transaction that it spends and the message signature
+// It requires as input the toSpend transaction that it spends
+// Transaction is build without any witness, so that the witness must be filled
+// by the caller.
 func GetToSignTx(toSpend *wire.MsgTx) *wire.MsgTx {
 	toSign := wire.NewMsgTx(toSignVersion)
 	toSign.LockTime = toSignLockTime
@@ -171,7 +173,7 @@ func Verify(
 	return vm.Execute()
 }
 
-func PubkeyToP2WPKHAddress(p *btcec.PublicKey, net *chaincfg.Params) (btcutil.Address, error) {
+func PubkeyToP2WPKHAddress(p *btcec.PublicKey, net *chaincfg.Params) (*btcutil.AddressWitnessPubKeyHash, error) {
 	witnessAddr, err := btcutil.NewAddressWitnessPubKeyHash(
 		btcutil.Hash160(p.SerializeCompressed()),
 		net,
@@ -184,7 +186,7 @@ func PubkeyToP2WPKHAddress(p *btcec.PublicKey, net *chaincfg.Params) (btcutil.Ad
 	return witnessAddr, nil
 }
 
-func PubKeyToP2TrSpendAddress(p *btcec.PublicKey, net *chaincfg.Params) (btcutil.Address, error) {
+func PubKeyToP2TrSpendAddress(p *btcec.PublicKey, net *chaincfg.Params) (*btcutil.AddressTaproot, error) {
 	tapKey := txscript.ComputeTaprootKeyNoScript(p)
 
 	address, err := btcutil.NewAddressTaproot(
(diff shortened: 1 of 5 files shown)


KonradStaniec:
@vitsalis do you happen to know what type of bip322 pop our frontend app generates ? I would guess it would be p2wpkh but better to be sure.

x/btcstaking/types/pop.go
```diff
@@ -77,6 +87,78 @@ func NewPoPWithECDSABTCSig(babylonSK cryptotypes.PrivKey, btcSK *btcec.PrivateKe
...
+	bip340PK := bbn.NewBIP340PubKeyFromBTCPK(btcPK)
+	babylonSig, err := babylonSK.Sign(*bip340PK)
+	if err != nil {
+		return nil, err
+	}
+	pop.BabylonSig = babylonSig
+
+	// TODO: temporary solution for MVP purposes.
+	// Eventually we need to use tmhash.Sum(pop.BabylonSig) rather than bbnSigHashHexBytes
+	// ref: https://github.com/babylonchain/babylon-private/issues/80
```
vitsalis, line 118: Let's migrate this issue and update the comment cc @SebastianElvis

x/btcstaking/types/pop.go
```diff
@@ -146,6 +228,64 @@ func (pop *ProofOfPossession) VerifyBIP340(babylonPK cryptotypes.PubKey, bip340P
...
 }
 
+func checkAddressAndWitness(address btcutil.Address, witness wire.TxWitness, net *chaincfg.Params) (CheckStakerKey, error) {
+	script, err := txscript.PayToAddrScript(address)
+
+	if err != nil {
+		return nil, err
+	}
+
+	// pay to taproot key spending path have only signaure in witness
```
vitsalis, line 238:
```suggestion
	// pay to taproot key spending path have only signature in witness
```

x/btcstaking/types/pop.go
```diff
@@ -146,6 +228,64 @@ func (pop *ProofOfPossession) VerifyBIP340(babylonPK cryptotypes.PubKey, bip340P
...
+
+			if err != nil {
+				return err
+			}
+
+			if bytes.Equal(keyAddress.ScriptAddress(), address.ScriptAddress()) {
+				return nil
+			} else {
+				return fmt.Errorf("bip322Sig.Address does not correspond to bip340PK")
+			}
```
vitsalis, line 257:
```suggestion
			if !bytes.Equal(keyAddress.ScriptAddress(), address.ScriptAddress()) {
				fmt.Errorf("bip322Sig.Address does not correspond to bip340PK")
			}
			return nil
```

x/btcstaking/types/pop.go
```diff
@@ -146,6 +228,64 @@ func (pop *ProofOfPossession) VerifyBIP340(babylonPK cryptotypes.PubKey, bip340P
...
+
+			if err != nil {
+				return err
+			}
+
+			if bytes.Equal(keyFromWitnessBytess, stakerKeyEncoded) {
+				return nil
+			} else {
+				return fmt.Errorf("bip322Sig.Address does not correspond to bip340PK")
+			}
```
vitsalis, line 282: Same as above

crypto/bip322/bip322.go
```diff
@@ -126,46 +117,46 @@ func GetToSpendTx(msg []byte, address string, net *chaincfg.Params) (*wire.MsgTx
 // GetToSignTx builds a toSign transaction based on the BIP-322 spec
 // https://github.com/bitcoin/bips/blob/e643d247c8bc086745f3031cdee0899803edea2f/bip-0322.mediawiki#full
 // It requires as input the toSpend transaction that it spends and the message signature
```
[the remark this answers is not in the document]
vitsalis: Let's also note that the witness is not included and should be filled in?

crypto/bip322/bip322.go
```diff
@@ -126,46 +117,46 @@ func GetToSpendTx(msg []byte, address string, net *chaincfg.Params) (*wire.MsgTx
...
+	address btcutil.Address,
+	net *chaincfg.Params) error {
 
-	toSign, err := GetToSignTx(toSpend, sig)
+	toSpend, err := GetToSpendTx(msg, address)
 	if err != nil {
 		return err
 	}
 
+	toSign := GetToSignTx(toSpend)
```
vitsalis, line 151: Maybe it's a good idea to pass the witness? In order to get the full `toSign` transaction, as it requires to have a witness.

crypto/bip322/bip322.go
```diff
@@ -179,3 +170,119 @@ func Verify(msg []byte, sig []byte, address string, net *chaincfg.Params) error
...
+		return nil, err
+	}
+	return address, nil
+}
+
+func SignWithP2WPKHAddress(
+	msg []byte,
+	privKey *btcec.PrivateKey,
+	net *chaincfg.Params,
+) (btcutil.Address, []byte, error) {
```
vitsalis, line 203: Shouldn't the return type be `*btcutil.AddressWitnessPubKeyHash` based on the returned types below?

crypto/bip322/bip322.go
```diff
@@ -179,3 +170,119 @@ func Verify(msg []byte, sig []byte, address string, net *chaincfg.Params) error
...
+	}
+
+	return witnessAddr, serializedWitness, nil
+}
+
+func SignWithP2TrSpendAddress(
+	msg []byte,
+	privKey *btcec.PrivateKey,
+	net *chaincfg.Params,
+) (btcutil.Address, []byte, error) {
```
vitsalis, line 248: Same here with `btcutil.AddressTaproot`

x/btcstaking/types/pop.go
```diff
@@ -162,14 +302,36 @@ func (pop *ProofOfPossession) VerifyBIP322(babylonPK cryptotypes.PubKey, bip340P
...
+	}
+
+	btcAddress, err := btcutil.DecodeAddress(bip322Sig.Address, net)
+
+	if err != nil {
+		return err
+	}
+
+	stakerKeyMatchesBtcAddressFn, err := checkAddressAndWitness(btcAddress, witness, net)
+
```
vitsalis, line 317: Maybe remove newlines between statements and checks of their error?

x/btcstaking/types/pop.go
```diff
@@ -178,7 +340,10 @@ func (pop *ProofOfPossession) VerifyBIP322(babylonPK cryptotypes.PubKey, bip340P
 		return fmt.Errorf("failed to verify pop.BabylonSig")
 	}
 
-	// TODO: rule 3: verify bip322Sig.Address corresponds to bip340PK
+	// rule 3: verify bip322Sig.Address corresponds to bip340PK
```
vitsalis, line 343: We could make this the first rule?

x/btcstaking/types/pop_test.go
```diff
@@ -97,3 +100,82 @@ func FuzzPoP_ECDSA(f *testing.F) {
...
+		// generate and verify PoP, correct case
+		pop, err := types.NewPoPWithBIP322P2TRBIP86Sig(babylonSK, btcSK, net)
+		require.NoError(t, err)
+		err = pop.VerifyBIP322(babylonPK, bip340PK, net)
+		require.NoError(t, err)
+	})
+}
+
+// TODO: Add more negative cases
+func TestValidBip322SigNotMatchingBip340PubKey(t *testing.T) {
```
vitsalis, line 151: Should we make this a fuzz test? Given that it uses randomness.

vitsalis (approved):
Looks good! Some questions and suggestions.

crypto/bip322/bip322.go
```diff
@@ -126,46 +117,46 @@ func GetToSpendTx(msg []byte, address string, net *chaincfg.Params) (*wire.MsgTx
...
+	toSign := GetToSignTx(toSpend)
+
+	toSign.TxIn[0].Witness = witness
+
 	// From the rules here:
 	// https://github.com/bitcoin/bips/blob/master/bip-0322.mediawiki#verification-process
 	// We only need to perform verification of whether toSign spends toSpend properly
 	// given that the signature is a simple one and we construct both toSpend and toSign
-	inputFetcher := txscript.NewCannedPrevOutputFetcher([]byte{}, 0)
+	inputFetcher := txscript.NewCannedPrevOutputFetcher(toSpend.TxOut[0].PkScript, 0)
```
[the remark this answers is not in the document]
KonradStaniec: So this is was actually a bug, as input fetcher should definitely return `toSpend` transaction pk script. I discovered it due to adding tests for taproot bip322 signing and verification.

My current guess, is that it make `txscript.NewTxSigHashes` to calculate and cache invalid transaction hashes and for some address types it did not matter and for some it make transaction signature invalid.

x/btcstaking/types/pop.go
```diff
@@ -9,11 +10,20 @@ import (
...
+	"github.com/btcsuite/btcd/wire"
 	"github.com/cometbft/cometbft/crypto/tmhash"
 	cryptotypes "github.com/cosmos/cosmos-sdk/crypto/types"
 )
 
+type CheckStakerKey func(stakerKey *bbn.BIP340PubKey) error
+
+type Bip322Sign func(sg []byte,
+	privKey *btcec.PrivateKey,
+	net *chaincfg.Params) (btcutil.Address, []byte, error)
```
SebastianElvis, line 25: Can these two functions be private or be defined in `crypto/bip322`? Either way it does not looke like something publicly accessible in `btcstaking/types`

crypto/bip322/bip322.go
KonradStaniec: hmm tbh mostly on BTC side, I have encoutared the flow:
1. create tx to send
2. create witness
3. fill in the witness in tx

This is mostly due to the fact that witness is a bit special field, in that regard that, unless requested directly witness is not part of transaction hash.

Also adding witness separately, makes it easier to share this part of code between signing and verifying bip322 signatures.

x/btcstaking/types/pop.go
```diff
@@ -146,6 +228,64 @@ func (pop *ProofOfPossession) VerifyBIP340(babylonPK cryptotypes.PubKey, bip340P
 	return nil
 }
 
+func checkAddressAndWitness(address btcutil.Address, witness wire.TxWitness, net *chaincfg.Params) (CheckStakerKey, error) {
```
SebastianElvis, line 231: same here, this is more like a part of `bip322` package rather than here

x/btcstaking/types/pop.go
```diff
@@ -162,14 +302,36 @@ func (pop *ProofOfPossession) VerifyBIP322(babylonPK cryptotypes.PubKey, bip340P
...
+		return err
+	}
+
+	btcAddress, err := btcutil.DecodeAddress(bip322Sig.Address, net)
+
+	if err != nil {
+		return err
+	}
+
+	stakerKeyMatchesBtcAddressFn, err := checkAddressAndWitness(btcAddress, witness, net)
```
SebastianElvis, line 316: Can we move L316-320 to somewhere like L343, such that we perform the cosnistency check of BTC address, witness and staker PK at a single place, and avoid retuning a function that assumes some external variables?

SebastianElvis (approved):
Great work! Some complimentary comments

x/btcstaking/types/pop.go
KonradStaniec: imo it is a bit safer to do this after sig verification, as signature verification excludes any possibility that witness is somehow maliciously crafted. As in one case, we parse public key from witness `(btcec.ParsePubKey(witness[1])` I would prefer to avoid situation in which this will panic or something. (low probability, but argument still holds, valid sig verifcation guarantess that bytes in the witness have expected shape)

x/btcstaking/types/pop_test.go
KonradStaniec: we can although it won't bring that much value imo, it will just always fail (although it is some kind of property to satisfy)

x/btcstaking/types/pop.go
KonradStaniec: yup they should be private. 

They should not be part of `crypto/bip322` imo, as in `crypto/bip322` we have concrete implementation and at pop level we just say `give me bip322 signing function (however it is implemented) and I will turn it into pop`

x/btcstaking/types/pop.go
KonradStaniec: imo not really, `bip322` do not care that much about address types and witness shapes. It says `give me address and witness, and I will tell you whether this witness satisfy this address`.

Only and pop level, we say that some types of witnesses and addresses are suitable to become proofs of possessions and some are not suitable.

x/btcstaking/types/pop.go
KonradStaniec: ahh of course we can, not sure why I had in my head that `checkAddressAndWitness` first 😅

KonradStaniec: ok I know why I structured it that way. 

I structured it this way to avoid situtuation in which we will get some big btc script to execute. As bip322 does not care about witness shape and address type, we could as well receive taproot script which executes multiple signatures verification i.e something like `OP_CHECKSIGADD .. OP_CHECKSIGADD.. OP_CHECKSIGADD..` this may be not correctly prices by our gas.

So that why my ideal flow for pop is:
1. parse data
2. verify address and witness are from the supported types
3. validate witness (bip322 verification)
4. check address matches our bip340 key

KonradStaniec: I think I will leave the flow as it is but add better doc for this decision.

crypto/bip322/bip322.go
```diff
@@ -179,3 +172,119 @@ func Verify(msg []byte, sig []byte, address string, net *chaincfg.Params) error
...
+	)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return witnessAddr, nil
+}
+
+func PubKeyToP2TrSpendAddress(p *btcec.PublicKey, net *chaincfg.Params) (*btcutil.AddressTaproot, error) {
```
jrwbabylonchain, line 189: @KonradStaniec 
Correct me if I'm wrong, both `PubkeyToP2WPKHAddress` and `PubKeyToP2TrSpendAddress` should be used to check if the public key is associated with the a address.
In that case, wouldn't it be better to have a generic method `PubKeyToAddress` where this method will try out both `PubKeyToP2TrSpendAddress` and `PubkeyToP2WPKHAddress` and return a type as well as the address? The returned type will indicate `p2wpkh, p2tr, p2pkh..` 

This would avoid the API to have knowledge about the different type BTC script/address, more importantly, if in the future babylon to support more BTC script/address types, the only place that need to make code change is babylon node.

KonradStaniec: > Correct me if I'm wrong, both PubkeyToP2WPKHAddress and PubKeyToP2TrSpendAddress should be used to check if the public key is associated with the a address.

Both them are pretty low level helpers (imo they should be in btcd), and they do not assume any particular usage. If you see how they are currently used, then only `PubKeyToP2TrSpendAddress` is used in way you mention, but `PubkeyToP2WPKHAddress` is only used when creating bip322 signatures, as with the data we have on babylon node there is better way to check association. 

> In that case, wouldn't it be better to have a generic method PubKeyToAddress where this method will try out both PubKeyToP2TrSpendAddress and PubkeyToP2WPKHAddress and return a type as well as the address? The returned type will indicate p2wpkh, p2tr, p2pkh..

I think I do not see how this `PubKeyToAddress` function would look like. It would need to take some kind of marker what address to create given that it is possible to create many address from on pub key. Could you maybe write how you see its function signature ?

PR #432, opened 2024-01-29
chore: Bump cosmos version with slow block fix

gitferry:
This PR does the following:
1. Bumped cosmos-sdk version which fixed the slow-block incident: https://github.com/cosmos/cosmos-sdk/commit/c4a2fe2b89875090a16b7c080f4e3015deaa6b9a
2. Injected extension signature to vote extension in the test helper due to (1). These changes are needed because (1) fixed a bug in `baseapp/abci_utils.go/ValidateVoteExtension` and we somehow bypassed the signature validation in our tests.
3. Cherry-picked some minor changes from the [PR](https://github.com/babylonchain/babylon-private/pull/160), mostly for cleanup stuff


x/zoneconcierge/keeper/ibc_header_decorator.go
```diff
@@ -75,9 +77,23 @@ func (d *IBCHeaderDecorator) PostHandle(ctx sdk.Context, tx sdk.Tx, simulate, su
...
+	// 2. If there is no error, then the cache context is written to the main context with no gas consumed
+	cacheCtx, write := ctx.CacheContext()
+	// CacheCtx's by default _share_ their gas meter with the parent.
+	// In our case, the cache ctx is given a new gas meter instance entirely,
+	// so gas usage is not counted towards tx gas usage.
+	//
+	// 50M is chosen as a large enough number to ensure that the posthandler will not run out of gas,
+	// but will eventually terminate in event of an accidental infinite loop with some gas usage.
+	upperGasLimitMeter := storetypes.NewGasMeter(storetypes.Gas(50_000_000))
+	cacheCtx = cacheCtx.WithGasMeter(upperGasLimitMeter)
```
SebastianElvis, line 92: This part of the code seems to be the deprecated one? Since right now we [execute](https://github.com/babylonchain/babylon/blob/4d1a193b0a0215ef3c7d430dbe051cff31b53c75/x/zoneconcierge/keeper/ibc_header_decorator.go#L68-L69) the msg in simulation mode as well, the gas estimation will be accurate. So it should be safe to use `ctx` here directly

app/test_helpers.go
```diff
@@ -59,6 +58,10 @@ func setup(t *testing.T, ps *PrivSigner, withGenesis bool, invCheckPeriod uint)
 	appOptions[flags.FlagHome] = nodeHome // ensure unique folder
 	appOptions[server.FlagInvCheckPeriod] = invCheckPeriod
 	appOptions["btc-config.network"] = string(bbn.BtcSimnet)
+	appOptions[server.FlagPruning] = pruningtypes.PruningOptionDefault
+	appOptions[server.FlagMempoolMaxTxs] = mempool.DefaultMaxTx
+	appOptions[flags.FlagChainID] = "chain-test"
+	baseAppOpts := server.DefaultBaseappOptions(appOptions)
```
SebastianElvis, line 64: Are these necessary?

SebastianElvis (approved):
Great work!

x/zoneconcierge/keeper/ibc_header_decorator.go
gitferry: This part of code is copied from the [PR](https://github.com/babylonchain/babylon-private/pull/160). Do you think we can revert the change of the whole file?

app/test_helpers.go
gitferry: These are needed to start the application with `SenderSequenceMempool` which is the same as the one in prod. Otherwise, it will use `NoOpMempool` in tests.

x/zoneconcierge/keeper/ibc_header_decorator.go
SebastianElvis: Yes. I think if you rebase the PR to the latest `dev` then the cache ctx thing will be gone, see https://github.com/babylonchain/babylon/blob/a606aebcbe8f7ebd6fb52c34b0dc61a92fe9c7e0/x/zoneconcierge/keeper/ibc_header_decorator.go#L64-L65

gitferry: I tried rebase but it didn't work. So I copy-paste the file from the latest dev to this PR. Please check whether it is expected. Thanks!

x/checkpointing/proposal.go
```diff
@@ -202,76 +202,66 @@ func (h *ProposalHandler) ProcessProposal() sdk.ProcessProposalHandler {
...
-		}
-		for _, txBytes := range req.Txs[1:] {
-			tx, err := h.txVerifier.ProcessProposalVerifyTx(txBytes)
+			var injectedCkpt ckpttypes.InjectedCheckpoint
+			if err := injectedCkpt.Unmarshal(injectedTx); err != nil {
+				h.logger.Error("failed to decode injected vote extension tx", "err", err)
+				// should not return error here as error will cause panic
+				return resReject, nil
+			}
+			req.Txs, err = removeInjectedTx(req.Txs)
```
vitsalis, line 219: Maybe add a note here that we remove the injected transaction just for the request, but the transaction will still be there in the block? This confused me a bit.

vitsalis (approved):
Nice work! Let's wait for @KonradStaniec approval as well

go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/btcsuite/btcd v0.23.5-0.20230711222809-7faa9b266231
 	github.com/cometbft/cometbft v0.38.5
 	github.com/cometbft/cometbft-db v0.9.1 // indirect
-	github.com/cosmos/cosmos-sdk v0.50.3
+	github.com/cosmos/cosmos-sdk v0.50.4-0.20240126152601-c4a2fe2b8987
```
KonradStaniec, line 10: unreleased sdk version ? Is this the one with the fix to our gasLimit problem ? If yes maybe leave a comment about that ?

x/checkpointing/proposal.go
```diff
@@ -202,76 +202,69 @@ func (h *ProposalHandler) ProcessProposal() sdk.ProcessProposalHandler {
...
+			//  and we should probably send an alarm and stall the blockchain
+			if !ckpt.Equal(injectedCkpt.Ckpt) {
+				// should not return error here as error will cause panic
+				h.logger.Error("invalid checkpoint in vote extension tx", "err", err)
+				return resReject, nil
 			}
 		}
 
+		// 5. verify the rest of the txs using the default handler
+		res, err := h.defaultHandler.ProcessProposalHandler()(ctx, req)
```
KonradStaniec, line 258: tbh I do not like that each time w create new handler i.e we call `ProcessProposalHandler` which return function and then we call the function.

Couldn't we create handlers one time, when wiring the app and then call them ?

x/checkpointing/proposal.go
```diff
@@ -311,9 +304,10 @@ func (h *ProposalHandler) PreBlocker() sdk.PreBlocker {
 	}
 }
 
-func getInjectedTx(txs [][]byte) ([]byte, error) {
-	if len(txs) == 0 {
-		return nil, fmt.Errorf("the proposal does not contain the checkpoint")
+// extractInjectedTx extracts the injected tx from the tx set
+func extractInjectedTx(txs [][]byte) ([]byte, error) {
```
KonradStaniec, line 308: couldn't this function return `InjectedCheckpoint` ?  (callers do it themselves anyway)

KonradStaniec (approved):
just minor code cleanliness comments, good work 👍

x/checkpointing/proposal.go
gitferry: Good point. Added `defaultPrepareHandler` and `defaultProcessHandler` to the `ProposalHandler` and wired them up creating the `ProposalHandler`

x/checkpointing/proposal.go
gitferry: Good point. Renamed `extractInjectedTx` to `extractInjectedCheckpoint` and added unmarshaling code in it so that the caller gets `InjectedCheckpoint` directly

PR #436, opened 2024-01-31
chore: fix flaky test due to incorrect bitmap bound

SebastianElvis:
`FuzzProofEpochSealed_BLSSig` was flaky because `GenFullBitmap` and `GenRandomBitmap` iterate over the entire bitmap to get validators in it. However, only the first 100 bits in the 104-bit bitmap are used and the rest shall not be counted. This PR fixes the flaky test by hardcoding the validator set size in `GenFullBitmap` and `GenRandomBitmap`. Fuzzed `FuzzProofEpochSealed_BLSSig` locally and the test is no longer flaky.


gitferry (approved):
Nice fix 👍

PR #440, opened 2024-02-01
metrics: add performance and application metrics for BTC staking

SebastianElvis:
Fixes [BM-1169](https://babylon-chain.atlassian.net/browse/BM-1169)

This PR introduces metrics for BTC staking. The metrics include both performance oriented ones and application specific ones. In particular, it includes

- *performance:* execution time of all relevant message handlers
- *BTC staking module:* number of finality providers and BTC delegations under each status, and total amount of staked Bitcoins
- *finality module:* last finalised height and last height (we can derive the gap of non finalised blocks by using them)

The implementation uses Cosmos SDK's native `telemetry` module.


filippos47 (approved):
Thanks, these will be super handy!
Out of curiosity, how did you choose from which entity to emit the metrics? Afaik, Prom metrics are exposed both from the Telemetry module and a dedicated Prom server (by default listens to `26660`).

x/btcstaking/keeper/keeper.go
```diff
@@ -78,28 +82,64 @@ func (k Keeper) BeginBlocker(ctx context.Context) error {
 	// for rewards
 	fpDistMap := map[string]*types.FinalityProviderDistInfo{}
 
+	// prepare metrics for {active, inactive} finality providers,
+	// {pending, active, unbonded BTC delegations}, and total staked Bitcoins
+	// NOTE: slashed finality providers and BTC delegations are recorded upon
+	// slashing events rather than here
+	var (
+		totalFPsMap          = map[string]struct{}{}
```
vitsalis, line 90: What's the expected memory overhead of this map?

x/btcstaking/keeper/keeper.go
```diff
@@ -78,28 +82,64 @@ func (k Keeper) BeginBlocker(ctx context.Context) error {
...
 
+	// record metrics for finality providers and total staked BTCs
+	numFPs := len(totalFPsMap)
+	numActiveFPs := numFPs
+	if numActiveFPs > int(params.MaxActiveFinalityProviders) {
+		numActiveFPs = int(params.MaxActiveFinalityProviders)
+	}
+	types.RecordActiveFinalityProviders(numActiveFPs)
+	types.RecordInactiveFinalityProviders(numFPs - numActiveFPs)
+	numStakedBTCs := float32(numStakedSats / SatoshisPerBTC)
```
vitsalis, line 136: Why do we want to convert to BTC? Satoshi is fine imo

x/btcstaking/keeper/keeper.go
```diff
@@ -78,28 +82,64 @@ func (k Keeper) BeginBlocker(ctx context.Context) error {
...
-			fpDistMap[fpBTCPKHex].AddBTCDel(btcDel, btcTipHeight, wValue, covenantQuorum)
+			fpDistMap[fpBTCPKHex].AddBTCDel(btcDel, btcTipHeight, wValue, params.CovenantQuorum)
 			return true
 		},
 	)
 
+	// record metrics for finality providers and total staked BTCs
+	numFPs := len(totalFPsMap)
+	numActiveFPs := numFPs
+	if numActiveFPs > int(params.MaxActiveFinalityProviders) {
```
vitsalis, line 131: From golang 1.21 there's a `max` function.

x/btcstaking/types/metrics.go
```diff
@@ -0,0 +1,103 @@
+package types
+
+import (
+	"github.com/cosmos/cosmos-sdk/telemetry"
+	"github.com/hashicorp/go-metrics"
+)
+
+// performance oriented metrics measuring the execution time of each message
```
vitsalis, line 8: Maybe metrics should be prefixed by the module name? e.g. `babylon/btcstaking/create_finality_provider`. 

Edit: I see below there's a label. Will this be enough? Is this what Cosmos SDK modules use?

x/btcstaking/types/metrics.go
```diff
@@ -0,0 +1,103 @@
...
+	// recording the number of slashed BTC delegations
+	MetricsKeyBTCDelegations = "btc_delegations"
+	// MetricsKeyStakedBitcoins is the key of the gauge recording the total
+	// amount of Bitcoins staked under active finality providers
+	MetricsKeyStakedBitcoins = "staked_bitcoins"
+)
+
+// RecordActiveFinalityProviders records the number of active finality providers.
+// It is triggered upon recording voting power table.
+func RecordActiveFinalityProviders(num int) {
```
vitsalis, line 34:
```suggestion
func RecordActiveFinalityProviders(num uint64) {
```

In general, should we change the other `int`s in this file to `uint64`s as well?

x/finality/types/metrics.go
```diff
@@ -0,0 +1,45 @@
...
+	// MetricsKeyLastHeight is the key of the gauge recording the last height
+	// of the ledger
+	MetricsKeyLastHeight = "last_height"
+	// MetricsKeyLastFinalizedHeight is the key of the gauge recording the
+	// last height finalized by finality providers
+	MetricsKeyLastFinalizedHeight = "last_finalized_height"
+)
+
+// RecordLastHeight records the last height. It is triggered upon `IndexBlock`
+func RecordLastHeight(height int) {
```
vitsalis, line 25: Same here for `int` vs. `uint64`.

vitsalis (commented):
This is very useful! Some questions about performance and types.

Also, let's change the base branch or wait until the new release before merging.

SebastianElvis:
> Afaik, Prom metrics are exposed both from the Telemetry module and a dedicated Prom server (by default listens to 26660).

I think all metrics specified using the `telemetry` package will be exposed by the prometheus server at :26660. The `telemetry` module gives all metrics specified by the developer to `"github.com/hashicorp/go-metrics"` (ref), more specifically to a singleton variable `globalMetrics`. When enabling metrics, the `Metrics` object will be created [here](https://github.com/cosmos/cosmos-sdk/blob/4e8d05b516c2c872577d283c45d6cf95cbc31f71/telemetry/metrics.go#L95-L159), which imports metrics from `"github.com/hashicorp/go-metrics"`. If you take a look at the code of Babylon and Cosmos SDK you might find that metrics specified by `telemetry` are presented in prometheus server.

x/btcstaking/keeper/finality_providers.go
```diff
@@ -53,6 +54,30 @@ func (k Keeper) SlashFinalityProvider(ctx context.Context, fpBTCPK []byte) error
...
 	fp.SlashedBtcHeight = btcTip.Height
 	k.SetFinalityProvider(ctx, fp)
+
+	// record newly slashed finality provider and BTC delegations
+	types.RecordNewSlashedFinalityProvider()
+	wValue := k.btccKeeper.GetParams(ctx).CheckpointFinalizationTimeout
+	covenantQuorum := k.GetParams(ctx).CovenantQuorum
+	btcDelIter := k.btcDelegatorStore(ctx, fp.BtcPk).Iterator(nil, nil)
+	defer btcDelIter.Close()
+	for ; btcDelIter.Valid(); btcDelIter.Next() {
```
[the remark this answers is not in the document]
SebastianElvis: This is indeed a valid concern. I cannot think of a way to record metrics for slashed BTC delegations with less than O(n) complexity, unless we index the number of active BTC delegations all the time.

One way of achieving this is to reuse the `RewardDistCache` which carries active BTC delegations. However, one needs to unmarshal such a big object and perform a search for the given finality provider. This seems not worth it either.

In light of this, how about we omit this metrics, but instead only record the number of slashed finality providers? For most scenarios the number of slashed finality providers seems good enough. If the consumer of the metrics is interested, it can perform a query to find out number of BTC delegations under this finality provider, and estimate how many of them are active. Wdyt?

x/btcstaking/keeper/keeper.go
SebastianElvis: Sats will be overflowed. Prometheus metrics only supports `float32` in the gauge type, see [here](https://github.com/babylonchain/babylon/blob/fa37be95eb61f5350c9cd70e4109dfba828d0076/x/btcstaking/types/metrics.go#L49-L53)

x/btcstaking/types/metrics.go
SebastianElvis: Yes Cosmos SDK uses module name as label to distinguish metrics between different modules. See usage [here](https://github.com/babylonchain/babylon/blob/fa37be95eb61f5350c9cd70e4109dfba828d0076/x/btcstaking/keeper/msg_server.go#L54-L55)

x/btcstaking/types/metrics.go
SebastianElvis: Given that this is the number of finality providers, probably `int` is more suitable. For example `len()` returns an int as well. Wdyt?

x/btcstaking/keeper/keeper.go
SebastianElvis: I just realised that a `int` is enough for this. Updated

SebastianElvis:
> Also, let's change the base branch or wait until the new release before merging.

Let's wait until the new release then. This PR will be affected by the new finality provider set rotation logic as well.

x/btcstaking/keeper/finality_providers.go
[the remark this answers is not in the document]
vitsalis: Yep sounds good with me

SebastianElvis:
Thanks for the comments! This PR is now rebased to the latest `dev` and revised according to comments. Please feel free to take a look again.

x/btcstaking/types/metrics.go
vitsalis: Hm no strong opinions here. My only concern is that we already know that these are unsigned integers so we could more strongly type this, but np either way.

x/btcstaking/keeper/keeper.go
```diff
@@ -97,15 +120,25 @@ func (k Keeper) BeginBlocker(ctx context.Context) error {
...
 			return true
 		},
 	)
+	// record metrics for finality providers and total staked BTCs
+	numActiveFPs := min(numFPs, int(params.MaxActiveFinalityProviders))
+	types.RecordActiveFinalityProviders(numActiveFPs)
+	types.RecordInactiveFinalityProviders(numFPs - numActiveFPs)
+	numStakedBTCs := float32(numStakedSats / SatoshisPerBTC)
+	types.RecordMetricsKeyStakedBitcoins(numStakedBTCs)
+	// record metrics for BTC delegations (NOTE: slashed BTC delegations are recorded upon slashing)
```
vitsalis, line 135: This is not true anymore?

x/btcstaking/types/metrics.go
```diff
@@ -0,0 +1,91 @@
...
+	keys := []string{MetricsKeyFinalityProviders, "INACTIVE"}
+	labels := []metrics.Label{telemetry.NewLabel(telemetry.MetricLabelNameModule, ModuleName)}
+	telemetry.SetGaugeWithLabels(
+		keys,
+		float32(num),
+		labels,
+	)
+}
+
+// RecordNewSlashedFinalityProvider increments the number slashed inactive finality providers.
```
vitsalis, line 56:
```suggestion
// RecordNewSlashedFinalityProvider increments the number of slashed finality providers.
```

vitsalis (approved):
Nice work!

x/btcstaking/keeper/keeper.go
SebastianElvis: Ah yes the note here needs to be removed. Done

PR #443, opened 2024-02-02
bench: benchmarking finality module

SebastianElvis:
This PR introduces benchmark tests for key operations in the finality module.

## Methodlogy

The goal is to identify performance bottlenecks and the corresponding issues in the normal operation of finality module, in a quick and dirty way. The PR suspects the following potential performance hotspots:

- committing randomness
- adding finality signatures
- tallying blocks

Thus this PR introduces 3 groups of benchmark tests:

- `BenchmarkCommitPubRandList`: repeatedly submits a given number of public randomness
- `BenchmarkAddFinalitySig`: repeatedly submits 1 finality signature
- `BenchmarkTallyBlocks`: repeatedly tallies blocks with a given number of finality providers

## Results

The overall result is that the above scenarios do not have performance issues.

### BenchmarkCommitPubRandList

```log
BenchmarkCommitPubRandList_10000-16          100         801347895 ns/op        160114001 B/op    331502 allocs/op
```

Committing 10000 public randomness costs 0.8s. That is, assuming 50 finality providers concurrently commit 2000 public randomness, then the execution time is 0.8s. This might be acceptable for now.

[commit-pub-rand.pdf](https://github.com/babylonchain/babylon/files/14135065/commit-pub-rand.pdf)

The profiling shows that operations that rely on iterators (`IsFirstPubRand` and `GetLastPubRand`) are hotspots. If we start optimising committing randomness then these could be the start.

### BenchmarkAddFinalitySig

```log
BenchmarkAddFinalitySig-16         77073            153013 ns/op            7082 B/op         50 allocs/op
```

Adding each finality signature takes only 0.00015s. So assuming we have 50 finality providers then each block will take 0.0075s to process all finality signatures. This is acceptable.

[add-finality-sig.pdf](https://github.com/babylonchain/babylon/files/14135112/add-finality-sig.pdf)

The profiling shows that verifying a EOTS signature is the hotspot. Verifying EOTS signature is inevitable, so optimising this might require optimising the EOTS implementation. No immediate actionable item here.

### BenchmarkTallyBlocks

```log
BenchmarkTallyBlocks_100-16          835           3484018 ns/op           43353 B/op        453 allocs/op
```

Tallying a block with 100 votes takes only 0.0035s. This is acceptable.

[tally.pdf](https://github.com/babylonchain/babylon/files/14135439/tally.pdf)

The profiling result excludes the execution overhead of tallying a block, meaning that tallying's execution overhead is negligible.

## TODOs

The benchmarking and profiling are not perfect because the BTC staking module is mocked. That means the overhead of accessing the database of BTC staking module is not considered in above results. Thus a TODO is to replace the mocked BTC staking module with a real one. Nevertheless, the results are expected to be similar with above, because only the finlaity providers and their voting power are accessed, and these are rather small.

### DIFF
--- a/x/btcstaking/keeper/bench_test.go
+++ b/x/btcstaking/keeper/bench_test.go
@@ -90,9 +90,8 @@ func benchBeginBlock(b *testing.B, numFPs int, numDelsUnderFP int) {
 	}
 }
 
-func BenchmarkBeginBlock_10_1(b *testing.B)  { benchBeginBlock(b, 10, 1) }
-func BenchmarkBeginBlock_10_10(b *testing.B) { benchBeginBlock(b, 10, 10) }
-
+func BenchmarkBeginBlock_10_1(b *testing.B)    { benchBeginBlock(b, 10, 1) }
+func BenchmarkBeginBlock_10_10(b *testing.B)   { benchBeginBlock(b, 10, 10) }
 func BenchmarkBeginBlock_10_100(b *testing.B)  { benchBeginBlock(b, 10, 100) }
 func BenchmarkBeginBlock_100_1(b *testing.B)   { benchBeginBlock(b, 100, 1) }
 func BenchmarkBeginBlock_100_10(b *testing.B)  { benchBeginBlock(b, 100, 10) }
--- a/x/finality/keeper/bench_test.go
+++ b/x/finality/keeper/bench_test.go
@@ -1,0 +1,69 @@
+package keeper_test
+
+import (
+	"fmt"
+	"math/rand"
+	"os"
+	"runtime/pprof"
+	"testing"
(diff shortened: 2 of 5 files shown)


gitferry (approved):
Awesome! One general question is that now we are only profiling CPU usage, do we also want to profile memory footprint?

SebastianElvis:
> Awesome! One general question is that now we are only profiling CPU usage, do we also want to profile memory footprint?

That's a good point. Currently if you execute the test with `-benchmem` flag it will measure the memory usage per operation (you can find the `7082 B/op` above). But yes we need to profile it further to find potential memory bottlenecks. Recorded https://github.com/babylonchain/babylon/issues/446

PR #447, opened 2024-02-05
btcstaking: optimise iterative access to BTC delegations

SebastianElvis:
This PR optimises the access on iterating all BTC delegations, including

- introducing `IterateActiveFPsAndBTCDelegations` that iterates over all BTC delagtions under non-slashed finality providers. It takes a handler function as input in order to be generic to any usage of iterating BTC delegations.
- merging the computation of voting power table and reward distribution cache by using `IterateActiveFPsAndBTCDelegations`. This reduces DB footprint of `BeginBlock` of BTC staking module by ~half.

This PR gives a speedup and CPU/memory foorprint reduction of 2x for BTC staking module's `BeginBlock`. After this PR, with 10k BTC delegations, `BeginBlock` takes ~0.006s in benchmarking tests.

```log
// current `dev`
BenchmarkBeginBlock_10_100-16               1707          12924721 ns/op         8420903 B/op     193448 allocs/op
// this PR
BenchmarkBeginBlock_10_100-16               4009          5902829 ns/op         3977645 B/op      90454 allocs/op
```

TODOs:
- [x] running benchmarks to verify performance improvment

### DIFF
--- a/x/btcstaking/keeper/keeper.go
+++ b/x/btcstaking/keeper/keeper.go
@@ -54,6 +54,11 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 	return ctx.Logger().With("module", fmt.Sprintf("x/%s", types.ModuleName))
 }
 
+// BeginBlocker is invoked upon `BeginBlock` of the system. The function
+// iterates over all BTC delegations under non-slashed finality providers
+// to 1) record the voting power table for the current height, and 2) record
+// the reward distribution cache used for distributing rewards once the block
+// is finalised by finality providers.
 func (k Keeper) BeginBlocker(ctx context.Context) error {
 	// index BTC height at the current height
 	k.IndexBTCHeight(ctx)
--- a/x/btcstaking/keeper/keeper.go
+++ b/x/btcstaking/keeper/keeper.go
@@ -75,13 +75,13 @@ func (k Keeper) BeginBlocker(ctx context.Context) error {
 
 	k.IterateActiveFPsAndBTCDelegations(
 		ctx,
-		func(fp *types.FinalityProvider, btcDel *types.BTCDelegation) {
+		func(fp *types.FinalityProvider, btcDel *types.BTCDelegation) bool {
 			fpBTCPKHex := fp.BtcPk.MarshalHex()
 
 			// record active finality providers
 			power := btcDel.VotingPower(btcTipHeight, wValue, covenantQuorum)
 			if power == 0 {
-				return // skip if no voting power
+				return true // skip if no voting power
 			}
 			fpPowerMap[fpBTCPKHex] += power
 
@@ -91,6 +91,7 @@ func (k Keeper) BeginBlocker(ctx context.Context) error {
 			}
 			// append BTC delegation
 			fpDistMap[fpBTCPKHex].AddBTCDel(btcDel, btcTipHeight, wValue, covenantQuorum)
+			return true
 		},
 	)
 
--- a/x/btcstaking/keeper/keeper.go
+++ b/x/btcstaking/keeper/keeper.go
@@ -1,8 +1,10 @@
 package keeper
 
 import (
-	corestoretypes "cosmossdk.io/core/store"
+	"context"
 	"fmt"
+
+	corestoretypes "cosmossdk.io/core/store"
 
 	"cosmossdk.io/log"
 	"github.com/babylonchain/babylon/x/btcstaking/types"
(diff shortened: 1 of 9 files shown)


x/btcstaking/keeper/voting_power_table.go
```diff
@@ -46,42 +36,44 @@ func (k Keeper) RecordVotingPowerTable(ctx context.Context) {
 			continue
 		}
 
-		fpPower := uint64(0)
-
 		// iterate all BTC delegations under this finality provider
 		// to calculate this finality provider's total voting power
 		btcDelIter := k.btcDelegatorStore(ctx, fpBTCPK).Iterator(nil, nil)
+		defer btcDelIter.Close()
```
vitsalis, line 42: Let's wrap this iteration in a function so that `defer` is called as soon as the execution finishes? e.g. as recommended [here](https://stackoverflow.com/a/45620423)

vitsalis (commented):
some initial stuff

x/btcstaking/keeper/voting_power_table.go
KonradStaniec, line 42: this will stack all defers closing  each `btcDelegatorStore` until the whole  ` IterateActiveFPsAndBTCDelegations` this means that individual stores will be open longer than necessary.

Maybe we can move it to separate function in which we will close each `btcDelegatorStore` right after iteration ?

x/btcstaking/keeper/keeper.go
```diff
@@ -51,3 +53,64 @@ func NewKeeper(
...
+	if len(fpPowerMap) == 0 {
+		return nil
+	}
+
+	// get top N finality providers and set their voting power to KV store
+	k.setCurrentTopNVotingPower(ctx, fpPowerMap)
+
+	// create reward distribution cache
+	rdc := types.NewRewardDistCache()
+	for _, fpDistInfo := range fpDistMap {
```
KonradStaniec, line 107: iteration over pointers (` fpDistInfo ` is pointer)using range is dangerus as you may end up passing only last pointer in iteration to `rdc.AddFinalityProviderDistInfo` 

lets make ` fpDistInfoCopy =  fpDistInfo` in next line and use this copy.

SebastianElvis: right, changed to iterate over keys, i.e.,  `for fpBTCPKHex := range fpDistMap`

x/btcstaking/keeper/keeper.go
```diff
@@ -51,3 +53,65 @@ func NewKeeper(
 func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 	return ctx.Logger().With("module", fmt.Sprintf("x/%s", types.ModuleName))
 }
+
+func (k Keeper) BeginBlocker(ctx context.Context) error {
```
vitsalis, line 57: docstring here would be useful

x/btcstaking/keeper/voting_power_table.go
```diff
@@ -47,45 +36,51 @@ func (k Keeper) RecordVotingPowerTable(ctx context.Context) {
...
 
+// setCurrentTopNVotingPower gets top N finality providers and set their current voting power to KV store
+func (k Keeper) setCurrentTopNVotingPower(ctx context.Context, fpPowerMap map[string]uint64) {
 	// filter out top `MaxActiveFinalityProviders` active finality providers in terms of voting power
+	activeFps := []*types.FinalityProviderWithMeta{}
+	for btcPKHex, power := range fpPowerMap {
+		btcPK, err := bbn.NewBIP340PubKeyFromHex(btcPKHex)
+		if err != nil {
+			panic(err) // only programming error
+		}
```
vitsalis, line 77: Previously we had a check of `power > 0`. We don't need it?

KonradStaniec (approved):
lgtm!

x/btcstaking/keeper/voting_power_table.go
SebastianElvis: You are right, we didn't. Put it here since 1) this signals a programming error or DB corruption which are still useful, and 2) it does not feel good to avoid handling an error :sweat_smile:

vitsalis: I'm not referring to the above error check, I'm referring to previously having [this](https://github.com/babylonchain/babylon/pull/447/files#diff-49e97a95c818e52cfd4bd1834e97fabe263acf941217ed10e81b153c6620b537L72) voting power check.

vitsalis (approved):

x/btcstaking/keeper/voting_power_table.go
SebastianElvis: Ah I see. We don't need this check anymore becuase [this](https://github.com/babylonchain/babylon/pull/447/files#diff-c53a7810ec5c6d9f2687565eed257f497247c84a8f226cba66eba692c2913a8aR86-R91) ensures that all entries in `fpPowerMap` will have positive voting power.

PR #448, opened 2024-02-05
fix: use defer for all iterators over KV store

SebastianElvis:


x/btcstaking/keeper/incentive.go
```diff
@@ -48,6 +48,7 @@ func (k Keeper) RecordRewardDistCache(ctx context.Context) {
 		// the finality provider's distribution info
 		fpDistInfo := types.NewFinalityProviderDistInfo(fp)
 		btcDelIter := k.btcDelegatorStore(ctx, fpBTCPK).Iterator(nil, nil)
+		defer btcDelIter.Close()
```
vitsalis, line 51: Recommend we wrap the whole `btcDelIter` into a function in order for the defered command to be executed as soon as this function wraps up. It can even be an anonymous function e.g. as recommended [here](https://stackoverflow.com/a/45620423)

x/btcstaking/keeper/voting_power_table.go
```diff
@@ -51,6 +52,7 @@ func (k Keeper) RecordVotingPowerTable(ctx context.Context) {
 		// iterate all BTC delegations under this finality provider
 		// to calculate this finality provider's total voting power
 		btcDelIter := k.btcDelegatorStore(ctx, fpBTCPK).Iterator(nil, nil)
+		defer btcDelIter.Close()
```
vitsalis, line 55: Same here

KonradStaniec, line 55: this will stack all defers closing  each `btcDelegatorStore` until the whole ` RecordVotingPowerTable` returns this means that individual stores will be open longer than necessary.

Maybe we can move it to separate function in which we will close each `btcDelegatorStore` right after iteration ? (the same comment goes to iteration in incentive)

vitsalis (approved):

PR #450, opened 2024-02-05
BTC timestamping phase 2 e2e hermes test

maurolacy:
Hermes relayer tests fixed by building hermes from a (currenlty) unmerged / unreleased hermes branch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -231,7 +231,7 @@ endif
 
 .PHONY: run-tests test test-all $(TEST_TARGETS)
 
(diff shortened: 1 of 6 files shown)


SebastianElvis (commented):
Given that the e2e test aims to ensure Hermes works with Babylon, the e2e test is supposed to use a released version of it. In light of this, maybe let's push everything with Go relayer for now, and do this PR until Hermes people release the version with that bugfix?

PR #451, opened 2024-02-05
Refactor adding covenant signatures

KonradStaniec:
Small non-breaking refactor of adding covenant signatures, also fixes some issues when processing covenant signatures message. Up until now:

- for every received covenant signatures message, we retrieved delegation from database two times: once at the beginning of processing message, and the second time when saving it to database (as our save was actually save-modify-write). 
Just to put that in context, for 100k delegation, this is 100k not necessary random database reads. Or actually more, as even  when we already reached quorum, and we were ignoring incoming covenant signatures, we still retrieved delegation second time from database just to not apply any changes to it.
- even though we ignored received signatures if quorum was already reached we emitted `&types.EventActivateBTCDelegation{BtcDel: btcDel}` event
- when quorum has already been reached or covenant member already voted we have gone through whole message processing, to just ignore all the results at the end.

### DIFF
--- a/x/btcstaking/keeper/msg_server.go
+++ b/x/btcstaking/keeper/msg_server.go
@@ -401,6 +401,16 @@ func (ms msgServer) AddCovenantSigs(goCtx context.Context, req *types.MsgAddCove
 		return nil, types.ErrInvalidCovenantPK.Wrapf("covenant pk: %s", req.Pk.MarshalHex())
 	}
 
+	if btcDel.IsSignedByCovMember(req.Pk) || btcDel.BtcUndelegation.IsSignedByCovMember(req.Pk) {
+		ms.Logger(ctx).Debug("Received duplicated covenant signature", "covenant pk", req.Pk.MarshalHex())
+		return &types.MsgAddCovenantSigsResponse{}, nil
+	}
+
+	if btcDel.HasCovenantQuorums(params.CovenantQuorum) {
+		ms.Logger(ctx).Debug("Received covenant signature after achieving quorum", "covenant pk", req.Pk.MarshalHex())
+		return &types.MsgAddCovenantSigsResponse{}, nil
+	}
+
 	// Note: we assume the order of adaptor sigs is matched to the
 	// order of finality providers in the delegation
 	// TODO: ensure the order for restaking, currently, we only have one finality provider
@@ -424,7 +434,7 @@ func (ms msgServer) AddCovenantSigs(goCtx context.Context, req *types.MsgAddCove
 		// this fails, it is a programming error
 		panic(err)
 	}
-	err = btcDel.SlashingTx.EncVerifyAdaptorSignatures(
+	parsedSlashingAdaptorSignatures, err := btcDel.SlashingTx.ParseEncVerifyAdaptorSignatures(
 		stakingInfo.StakingOutput,
 		slashingSpendInfo,
 		req.Pk,
@@ -485,7 +495,7 @@ func (ms msgServer) AddCovenantSigs(goCtx context.Context, req *types.MsgAddCove
 		// this fails, it is a programming error
 		panic(err)
 	}
-	err = btcDel.BtcUndelegation.SlashingTx.EncVerifyAdaptorSignatures(
+	parsedUnbondingSlashingAdaptorSignatures, err := btcDel.BtcUndelegation.SlashingTx.ParseEncVerifyAdaptorSignatures(
 		unbondingOutput,
 		unbondingSlashingSpendInfo,
 		req.Pk,
@@ -496,17 +506,11 @@ func (ms msgServer) AddCovenantSigs(goCtx context.Context, req *types.MsgAddCove
 		return nil, types.ErrInvalidCovenantSig.Wrapf("err: %v", err)
 	}
 
-	// all good, add signatures to BTC delegation/undelegation and set it back to KVStore
-	if err := ms.AddCovenantSigsToBTCDelegation(
-		ctx,
-		req.StakingTxHash,
-		req.Pk,
-		req.SlashingTxSigs,
-		req.UnbondingTxSig,
-		req.SlashingUnbondingTxSigs,
-	); err != nil {
-		return nil, types.ErrInvalidCovenantSig.Wrapf("failed to add covenant signatures to BTC delegation: %v", err)
-	}
+	// All is fine add recieved signatures to the BTC delegation and BtcUndelegation
+	btcDel.AddCovenantSigs(req.Pk, parsedSlashingAdaptorSignatures, params.CovenantQuorum)
+	btcDel.BtcUndelegation.AddCovenantSigs(req.Pk, req.UnbondingTxSig, parsedUnbondingSlashingAdaptorSignatures, params.CovenantQuorum)
+
+	ms.setBTCDelegation(ctx, btcDel)
 
 	// notify subscriber
 	if err := ctx.EventManager().EmitTypedEvent(&types.EventActivateBTCDelegation{BtcDel: btcDel}); err != nil {
--- a/x/btcstaking/keeper/msg_server.go
+++ b/x/btcstaking/keeper/msg_server.go
@@ -401,7 +401,7 @@ func (ms msgServer) AddCovenantSigs(goCtx context.Context, req *types.MsgAddCove
 		return nil, types.ErrInvalidCovenantPK.Wrapf("covenant pk: %s", req.Pk.MarshalHex())
 	}
 
-	if btcDel.IsSignedByCovMember(req.Pk) || btcDel.BtcUndelegation.IsSignedByCovMember(req.Pk) {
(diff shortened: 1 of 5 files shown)


x/btcstaking/keeper/msg_server.go
```diff
@@ -401,6 +401,16 @@ func (ms msgServer) AddCovenantSigs(goCtx context.Context, req *types.MsgAddCove
 		return nil, types.ErrInvalidCovenantPK.Wrapf("covenant pk: %s", req.Pk.MarshalHex())
 	}
 
+	if btcDel.IsSignedByCovMember(req.Pk) || btcDel.BtcUndelegation.IsSignedByCovMember(req.Pk) {
```
SebastianElvis, line 404: Should this be

```suggestion
	if btcDel.IsSignedByCovMember(req.Pk) && btcDel.BtcUndelegation.IsSignedByCovMember(req.Pk) {
```

which is the expected behaviour of a covenant member. If the covenant member only submits one of these signatures then the message will be rejected anyway

x/btcstaking/keeper/msg_server.go
```diff
@@ -496,17 +506,11 @@ func (ms msgServer) AddCovenantSigs(goCtx context.Context, req *types.MsgAddCove
...
-		req.UnbondingTxSig,
-		req.SlashingUnbondingTxSigs,
-	); err != nil {
-		return nil, types.ErrInvalidCovenantSig.Wrapf("failed to add covenant signatures to BTC delegation: %v", err)
-	}
+	// All is fine add recieved signatures to the BTC delegation and BtcUndelegation
+	btcDel.AddCovenantSigs(req.Pk, parsedSlashingAdaptorSignatures, params.CovenantQuorum)
+	btcDel.BtcUndelegation.AddCovenantSigs(req.Pk, req.UnbondingTxSig, parsedUnbondingSlashingAdaptorSignatures, params.CovenantQuorum)
+
+	ms.setBTCDelegation(ctx, btcDel)
```
SebastianElvis, line 513: Shall we use `updateBTCDelegation` to update a BTC delegation instead? Alternatively we can remove `updateBTCDelegation` function given that it's not frequently used.

x/btcstaking/types/btc_slashing_tx.go
```diff
@@ -193,25 +193,28 @@ func (tx *BTCSlashingTx) EncVerifyAdaptorSignature(
...
 	slashingSpendInfo *btcstaking.SpendInfo,
 	pk *bbn.BIP340PubKey,
 	valPKs []bbn.BIP340PubKey,
 	sigs [][]byte,
-) error {
-	for i, sig := range sigs {
+) ([]asig.AdaptorSignature, error) {
+	var adaptorSigs []asig.AdaptorSignature = make([]asig.AdaptorSignature, len(sigs))
+	for i, s := range sigs {
+		sig := s
```
SebastianElvis, line 210:
```suggestion
	for i := range sigs {
		sig := sigs[i]
```
Probably a nitpick but the above seems more natural

x/btcstaking/types/btc_slashing_tx.go
```diff
@@ -222,10 +225,32 @@ func (tx *BTCSlashingTx) EncVerifyAdaptorSignatures(
...
+		adaptorSigs[i] = *adaptorSig
+	}
+	return adaptorSigs, nil
+}
+
+// EncVerifyAdaptorSignatures verifies a list of adaptor signatures, each
+// encrypted by a restaked validator PK and signed by the given PK, w.r.t. the
+// given funding output (in staking or unbonding tx), slashing spend info and
+// slashing tx
+func (tx *BTCSlashingTx) EncVerifyAdaptorSignatures(
```
SebastianElvis, line 239: Is this function still necessary? looks like it's doing the same thing as `ParseEncVerifyAdaptorSignatures`

x/btcstaking/keeper/msg_server.go
```diff
@@ -496,17 +506,11 @@ func (ms msgServer) AddCovenantSigs(goCtx context.Context, req *types.MsgAddCove
...
-		req.Pk,
-		req.SlashingTxSigs,
-		req.UnbondingTxSig,
-		req.SlashingUnbondingTxSigs,
-	); err != nil {
-		return nil, types.ErrInvalidCovenantSig.Wrapf("failed to add covenant signatures to BTC delegation: %v", err)
-	}
+	// All is fine add recieved signatures to the BTC delegation and BtcUndelegation
+	btcDel.AddCovenantSigs(req.Pk, parsedSlashingAdaptorSignatures, params.CovenantQuorum)
+	btcDel.BtcUndelegation.AddCovenantSigs(req.Pk, req.UnbondingTxSig, parsedUnbondingSlashingAdaptorSignatures, params.CovenantQuorum)
```
SebastianElvis, line 511: Given that we always add covenant sigs to both delegation and undelegation, probably we can make merge the two functions?

x/btcstaking/types/btc_delegation.go
```diff
@@ -198,25 +198,14 @@ func (d *BTCDelegation) IsSignedByCovMember(covPk *bbn.BIP340PubKey) bool {
...
-	// we can ignore the covenant sig if quorum is already reached
-	if d.HasCovenantQuorums(quorum) {
-		return nil
-	}
-
-	if d.IsSignedByCovMember(covPk) {
-		return nil
-	}
-
+func (d *BTCDelegation) AddCovenantSigs(covPk *bbn.BIP340PubKey, sigs []asig.AdaptorSignature, quorum uint32) {
```
SebastianElvis, line 201:
```suggestion
func (d *BTCDelegation) AddCovenantSigs(covPk *bbn.BIP340PubKey, sigs []asig.AdaptorSignature) {
```
Quorum is not verified in this function anymore

x/btcstaking/types/btc_undelegation.go
```diff
@@ -72,16 +72,7 @@ func (ud *BTCUndelegation) AddCovenantSigs(
 	unbondingSig *bbn.BIP340Signature,
 	slashingSigs []asig.AdaptorSignature,
 	quorum uint32,
```
SebastianElvis, line 74:
```suggestion
	unbondingSig *bbn.BIP340Signature,
	slashingSigs []asig.AdaptorSignature,
```

x/btcstaking/types/btc_delegation.go
SebastianElvis: Also this modification seems to add a contract that `AddCovenantSigs` shall not take duplicated covenant adaptor signatrues. While it might be acceptable given that `BTCDelegation` will be entirely a internal database object, maybe it's a good idea to document these contracts in the docstring. Wdyt?

SebastianElvis (commented):
Much cleaner!

x/btcstaking/keeper/msg_server.go
KonradStaniec: tbh I do not see much difference between using either `&&` or `||`, adding both types of signatures should be atomic so in theory it would be enough to check only `btcDel.IsSignedByCovMember(req.Pk)` . 

This makes me think that there can be more refactoring done here. 

I will change it to`&&` for now as this is maybe more complete.

x/btcstaking/keeper/msg_server.go
KonradStaniec: I think we can gradually remove `updateBTCDelegation` as it does this `update-modify-write` operations, and in most staking handlers we already have delegation in memory. 

I think it is used only in one other place, when receving unbonding signature. I will remove it form there and remove it entirely.

x/btcstaking/types/btc_slashing_tx.go
KonradStaniec: It is public, and I do not want to break other programs (maybe covenant it is using it). We may consider removing it after testnet.

x/btcstaking/types/btc_delegation.go
KonradStaniec: hum I just see those two methods adding signatures as little helper to mutate object state, although I agres that it is good idea to add some doc string that it it up to the caller to do all verifications 👍

x/btcstaking/types/btc_slashing_tx.go
```diff
@@ -193,25 +193,28 @@ func (tx *BTCSlashingTx) EncVerifyAdaptorSignature(
...
 }
 
-// EncVerifyAdaptorSignatures verifies a list of adaptor signatures, each
+// ParseEncVerifyAdaptorSignatures verifies a list of adaptor signatures, each
 // encrypted by a restaked validator PK and signed by the given PK, w.r.t. the
 // given funding output (in staking or unbonding tx), slashing spend info and
 // slashing tx
-func (tx *BTCSlashingTx) EncVerifyAdaptorSignatures(
+// It returns a list of parsed adaptor signatures in case of successful verification
+func (tx *BTCSlashingTx) ParseEncVerifyAdaptorSignatures(
```
SebastianElvis, line 201: Probably update the docstring to say this function returns parsed adaptor signatures?

SebastianElvis (approved):
Always nice to see PRs that remove dead code :clap:

PR #453, opened 2024-02-06
fix non determinism in `BeginBlock`

SebastianElvis:
There is a non determinism due to the usage of [map](https://github.com/babylonchain/babylon/commit/6f60ff0e608271f01fdb0e8072deafa736c8c951#diff-c53a7810ec5c6d9f2687565eed257f497247c84a8f226cba66eba692c2913a8aR73-R79). This PR fixes it by avoiding using maps. It breaks down the nested iterators into two functions and handles each finality provider sequentially. Only array is used.

### DIFF
--- a/x/btcstaking/keeper/btc_delegators.go
+++ b/x/btcstaking/keeper/btc_delegators.go
@@ -54,6 +54,29 @@ func (k Keeper) AddBTCDelegation(ctx context.Context, btcDel *types.BTCDelegatio
 	return nil
 }
 
+// IterateBTCDelegations iterates all BTC delegations under a given finality provider
+func (k Keeper) IterateBTCDelegations(ctx context.Context, fpBTCPK *bbn.BIP340PubKey, handler func(btcDel *types.BTCDelegation) bool) {
+	btcDelIter := k.btcDelegatorStore(ctx, fpBTCPK).Iterator(nil, nil)
+	defer btcDelIter.Close()
+	for ; btcDelIter.Valid(); btcDelIter.Next() {
+		// unmarshal delegator's delegation index
+		var btcDelIndex types.BTCDelegatorDelegationIndex
+		k.cdc.MustUnmarshal(btcDelIter.Value(), &btcDelIndex)
+		// retrieve and process each of the BTC delegation
+		for _, stakingTxHashBytes := range btcDelIndex.StakingTxHashList {
+			stakingTxHash, err := chainhash.NewHash(stakingTxHashBytes)
+			if err != nil {
+				panic(err) // only programming error is possible
+			}
+			btcDel := k.getBTCDelegation(ctx, *stakingTxHash)
+			shouldContinue := handler(btcDel)
+			if !shouldContinue {
+				return
+			}
+		}
+	}
+}
+
 // hasBTCDelegatorDelegations checks if the given BTC delegator has any BTC delegations under a given finality provider
 func (k Keeper) hasBTCDelegatorDelegations(ctx context.Context, fpBTCPK *bbn.BIP340PubKey, delBTCPK *bbn.BIP340PubKey) bool {
 	fpBTCPKBytes := fpBTCPK.MustMarshal()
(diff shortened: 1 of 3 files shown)


x/btcstaking/keeper/voting_power_table.go
```diff
@@ -13,19 +12,13 @@ import (
...
 	fpIter := k.finalityProviderStore(ctx).Iterator(nil, nil)
 	defer fpIter.Close()
 	for ; fpIter.Valid(); fpIter.Next() {
 		fpBTCPKBytes := fpIter.Key()
-		fpBTCPK, err := bbn.NewBIP340PubKey(fpBTCPKBytes)
-		if err != nil {
-			// failed to unmarshal finality provider PK in KVStore is a programming error
-			panic(err)
-		}
 		fp, err := k.GetFinalityProvider(ctx, fpBTCPKBytes)
```
KonradStaniec, line 22: isn't this actually `fpIter.Value` ? i.e instead of doing database call you could do unmarshal `fpIter.Value()`

x/btcstaking/keeper/keeper.go
```diff
@@ -71,51 +71,55 @@ func (k Keeper) BeginBlocker(ctx context.Context) error {
...
 		ctx,
-		func(fp *types.FinalityProvider, btcDel *types.BTCDelegation) bool {
-			fpBTCPKHex := fp.BtcPk.MarshalHex()
-
-			// record active finality providers
-			power := btcDel.VotingPower(btcTipHeight, wValue, covenantQuorum)
-			if power == 0 {
-				return true // skip if no voting power
+		func(fp *types.FinalityProvider) bool {
+			fpPower := uint64(0)
```
KonradStaniec, line 84: I guess this is not really necessary as `fpDistInfo` contains total power

SebastianElvis: right, got rid of all fpPower things and cut some more code

x/btcstaking/keeper/keeper.go
```diff
@@ -71,51 +71,49 @@ func (k Keeper) BeginBlocker(ctx context.Context) error {
...
-			power := btcDel.VotingPower(btcTipHeight, wValue, covenantQuorum)
-			if power == 0 {
-				return true // skip if no voting power
+		func(fp *types.FinalityProvider) bool {
+			fpDistInfo := types.NewFinalityProviderDistInfo(fp)
+
+			// iterate over all BTC delegations under the finality provider
+			// in order to accumulate voting power and reward dist info for it
+			k.IterateBTCDelegations(ctx, fp.BtcPk, func(btcDel *types.BTCDelegation) bool {
+				// accumulate voting power and reward distirbution cache
```
gitferry, line 89:
```suggestion
				// accumulate voting power and reward distribution cache
```

x/btcstaking/keeper/btc_delegators.go
```diff
@@ -54,6 +54,29 @@ func (k Keeper) AddBTCDelegation(ctx context.Context, btcDel *types.BTCDelegatio
...
+		var btcDelIndex types.BTCDelegatorDelegationIndex
+		k.cdc.MustUnmarshal(btcDelIter.Value(), &btcDelIndex)
+		// retrieve and process each of the BTC delegation
+		for _, stakingTxHashBytes := range btcDelIndex.StakingTxHashList {
+			stakingTxHash, err := chainhash.NewHash(stakingTxHashBytes)
+			if err != nil {
+				panic(err) // only programming error is possible
+			}
+			btcDel := k.getBTCDelegation(ctx, *stakingTxHash)
+			shouldContinue := handler(btcDel)
```
gitferry, line 72: I don't see any cases the handler will return `false`. Maybe we can just remove the return value of the handler? Also, it confused me that the handler would return a boolean. Maybe it can return an error.

gitferry (approved):
Lgtm! Minor comments:

x/btcstaking/keeper/btc_delegators.go
SebastianElvis: The returned value is [reserved](https://github.com/babylonchain/babylon/pull/447#discussion_r1477790231) for other use cases when the code does not want to iterate the rest ones anymore

KonradStaniec (approved):

PR #456, opened 2024-02-07
fix: fix flaky test `FuzzWithdrawReward`

SebastianElvis:
`FuzzWithdrawReward` was [flaky](https://app.circleci.com/pipelines/github/babylonchain/babylon/2019/workflows/6b2a669d-867e-446e-a218-f7d49b7090df/jobs/3399) because `amount` [here](https://github.com/babylonchain/babylon/blob/396a712297f97cd8ca225bcea81ee9b02dada991/testutil/datagen/incentive.go#L65-L66) might be 1, making `RandomInt` panic. This PR fixes it by adding a check. Fuzzed the test locally and the fix works.


gitferry (approved):
Nice!

PR #457, opened 2024-02-07
btcstaking: unify BTC delegation state update events and distribution cache

SebastianElvis:
Fixes [BM-1204](https://babylon-chain.atlassian.net/browse/BM-1204) (step 1 of [BM-1195](https://babylon-chain.atlassian.net/browse/BM-1195))

This PR does some preliminary cleanup tasks for preparing the implementation of improved voting power distribution calculation, including:

- unifying `EventNewBTCDelegation`, `EventActivateBTCDelegation`, and `EventUnbondedBTCDelegation` into a single generic one `EventBTCDelegationStateUpdate` that is emitted whenever BTC delegation state is updated. It also cuts unnecessary fields to save the block space.
- extending `BTCDelDistInfo` to also have BTC PK. This allows to reconcile all state update events with the reward distribution cache to get the new reward distribution cache (which will be done in subsequent PRs).
- reusing reward distribution cache for recording the voting power table, and renaming `RewardDistCache` to `VotingPowerDistCache`. This unifies the processing of voting power table and reward distribution cache.


x/btcstaking/keeper/keeper.go
```diff
@@ -70,10 +70,8 @@ func (k Keeper) BeginBlocker(ctx context.Context) error {
 	}
 	wValue := k.btccKeeper.GetParams(ctx).CheckpointFinalizationTimeout
 
-	// prepare for recording finality providers with positive voting power
-	activeFps := []*types.FinalityProviderWithMeta{}
 	// prepare for recording finality providers and their BTC delegations
-	// for rewards
+	// for voting power and rewards
 	rdc := types.NewRewardDistCache()
```
gitferry, line 75:
```suggestion
	rdc := types.NewVotingPowerDistCache()
```
I just feel like this is essentially what this cache stores. Or we can rename it something similar.

x/btcstaking/keeper/msg_server.go
```diff
@@ -411,6 +415,14 @@ func (ms msgServer) AddCovenantSigs(goCtx context.Context, req *types.MsgAddCove
 		return &types.MsgAddCovenantSigsResponse{}, nil
 	}
 
+	// ensure BTC delegation is still pending, i.e., not expired
+	btcTipHeight := ms.btclcKeeper.GetTipInfo(ctx).Height
+	wValue := ms.btccKeeper.GetParams(ctx).CheckpointFinalizationTimeout
+	status := btcDel.GetStatus(btcTipHeight, wValue, params.CovenantQuorum)
+	if status != types.BTCDelegationStatus_PENDING {
+		return nil, types.ErrInvalidDelegationState.Wrapf("the BTC delegation is expected to be %s, but is %s", types.BTCDelegationStatus_PENDING.String(), status.String())
```
gitferry, line 423:
```suggestion
		return nil, types.ErrInvalidDelegationState.Wrapf("expected: %s, got: %s", types.BTCDelegationStatus_PENDING.String(), status.String())
```

gitferry (approved):
The separation of tasks is very clean! Only some minor comments:

x/btcstaking/keeper/keeper.go
SebastianElvis: Good point. I was also wondering whether to update the name since the struct is also used for voting power table now. Renamed `RewardDistCache` to `VotingPowerDistCache` :+1:

PR #458, opened 2024-02-07
Add CHANGELOG

maurolacy:
Wrapper around `github_changelog_generator` that simplifies keeping the changelog updated.

To use it, you first need to install https://github.com/github-changelog-generator/github-changelog-generator of course.

Also, though it can be used without it, it's convenient to have a token called `CHANGELOG_GITHUB_TOKEN`, to increase the number of API requests that can you do when generating the changelog entries.

See https://github.com/github-changelog-generator/github-changelog-generator?tab=readme-ov-file#github-token for instructions on how to generate and set the token.

A convenient way to use it, is with the `-u` (upcoming) flag, to pass a version tag that that is not existing yet in the repo.
That way you can update the changelog in a release branch, before tagging the new version. And the script takes care of creating the new header / section for you.

Finally, nothing prevents you from editing a section of the changelog manually. As long as the section headers are kept consistent, the script will continue to work.

Also, you can of course still call `github_changelog_generator` directly instead. By example for passing extra parameters / setting extra flags.
The ideal thing to do here would be to add those parameters (as a passthrough) to the script, so, if you find a useful flag / parameter that you want / need to use, let me know and I'll add it to the wrapper script.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [v0.8.0](https://github.com/babylonchain/babylon/tree/v0.8.0) (2024-02-08)
+
+[Full Changelog](https://github.com/babylonchain/babylon/compare/v0.8.0-rc.0...v0.8.0)
+
 ## [v0.8.0-rc.0](https://github.com/babylonchain/babylon/tree/v0.8.0-rc.0) (2024-01-22)
 
 [Full Changelog](https://github.com/babylonchain/babylon/compare/v0.7.2...v0.8.0-rc.0)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,1 +1,526 @@
 # Changelog
+
+## [v0.8.0-rc.0](https://github.com/babylonchain/babylon/tree/v0.8.0-rc.0) (2024-01-22)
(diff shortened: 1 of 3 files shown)


SebastianElvis (approved):
This is super helpful!

One minor comment: maybe it's good to add `Makefile` entries for using this command?

gitferry (approved):
Great work! One suggestion of formatting the CHANGELOG: maybe we can introduce some tool to enforce the linter of the PR title, e.g., https://github.com/dreampulse/action-lint-pull-request-title?tab=readme-ov-file. Generally, a PR should not be merged if the linter does not pass.

maurolacy:
> Great work! One suggestion of formatting the CHANGELOG: maybe we can introduce some tool to enforce the linter of the PR title, e.g., https://github.com/dreampulse/action-lint-pull-request-title?tab=readme-ov-file. Generally, a PR should not be merged if the linter does not pass.

Do that as part of PR creation I guess? Will check it, and we can add it.

maurolacy:
> One minor comment: maybe it's good to add `Makefile` entries for using this command?

Added. You can use it like `make update-changelog SINCE_TAG=v0.8.0`.

PR #461, opened 2024-02-09
btcstaking: efficient voting power distribution update

SebastianElvis:
This PR provides the full implementation of the efficient voting power distribution update algorithm, where we compare the last voting power table and the staking-related events at the current height to derive the current voting power table. 

This PR consists of a series of PRs below that have been reviewed, and has been rebased to the latest `dev`. The overall change needs to be reviewed again. It needs to be merged via a merge commit.


x/btcstaking/keeper/finality_providers.go
```diff
@@ -39,13 +38,18 @@ func (k Keeper) GetFinalityProvider(ctx context.Context, fpBTCPK []byte) (*types
...
+
+	// ensure finality provider is not slashed yet
 	if fp.IsSlashed() {
 		return types.ErrFpAlreadySlashed
 	}
+
+	// set finality provider to be slashed
 	fp.SlashedBabylonHeight = uint64(sdk.UnwrapSDKContext(ctx).HeaderInfo().Height)
 	btcTip := k.btclcKeeper.GetTipInfo(ctx)
 	if btcTip == nil {
```
gitferry, line 55: maybe we can just return error here and let the caller decide whether to panic (the two callers will panic upon errors anyway)

x/btcstaking/keeper/grpc_query.go
```diff
@@ -222,10 +222,7 @@ func (k Keeper) FinalityProviderDelegations(ctx context.Context, req *types.Quer
 			return err
 		}
 
-		curBTCDels, err := k.getBTCDelegatorDelegations(sdkCtx, fpPK, delBTCPK)
-		if err != nil {
-			return err
-		}
+		curBTCDels := k.getBTCDelegatorDelegations(sdkCtx, fpPK, delBTCPK)
```
gitferry, line 225: should we do nil check here?

x/btcstaking/keeper/msg_server.go
```diff
@@ -457,6 +453,14 @@ func (ms msgServer) AddCovenantSigs(goCtx context.Context, req *types.MsgAddCove
...
 	}
 
+	// ensure BTC delegation is still pending, i.e., not expired
+	btcTipHeight := ms.btclcKeeper.GetTipInfo(ctx).Height
+	wValue := ms.btccKeeper.GetParams(ctx).CheckpointFinalizationTimeout
+	status := btcDel.GetStatus(btcTipHeight, wValue, params.CovenantQuorum)
+	if status != types.BTCDelegationStatus_PENDING {
+		return nil, types.ErrInvalidDelegationState.Wrapf("expected: %s, got: %s", types.BTCDelegationStatus_PENDING.String(), status.String())
+	}
+
```
gitferry, line 463: This check seems unnecessary and could complicate covenant-emulator's error handling, similar to the case of duplicate covenant sig (already changed to returning nil).

x/btcstaking/keeper/power_dist_change.go
```diff
@@ -0,0 +1,317 @@
...
+	types.RecordInactiveFinalityProviders(numInactiveFPs)
+	// staked Satoshi
+	// TODO: some wrapper functions between Satoshis and voting power
+	// to make the 1:1 conversion clear
+	stakedSats := uint64(0)
+	for _, fp := range dc.FinalityProviders {
+		stakedSats += fp.TotalVotingPower
+	}
+	numStakedBTCs := float32(stakedSats / SatoshisPerBTC)
+	types.RecordMetricsKeyStakedBitcoins(numStakedBTCs)
```
gitferry, line 90:
```suggestion
	stakedSats := btcutil.Amount(0)
	for _, fp := range dc.FinalityProviders {
		// 1 unit of voting power equals to 1 sat
		powerToSats := btcutil.Amount(fp.TotalVotingPower)
		stakedSats += powerToSats
	}
	numStakedBTCs := stakedSats.ToBTC()
	types.RecordMetricsKeyStakedBitcoins(float32(numStakedBTCs))
```
Just wanted to kill this TODO 😄

gitferry (approved):
Great work! Some minor comments:

x/btcstaking/keeper/power_dist_change.go
```diff
@@ -0,0 +1,317 @@
...
+// This is called after processing all BTC delegation events in `BeginBlocker`
+// nolint:unused
+func (k Keeper) ClearPowerDistUpdateEvents(ctx context.Context, btcHeight uint64) {
+	store := k.powerDistUpdateEventStore(ctx, btcHeight)
+	keys := [][]byte{}
+
+	// get all keys
+	// using an enclosure to ensure iterator is closed right after
+	// the function is done
+	func() {
```
KonradStaniec, line 261: maybe move it to separate function like `GetAllEventKeys` ?

x/btcstaking/keeper/power_dist_change.go
```diff
@@ -0,0 +1,317 @@
...
+func (k Keeper) addPowerDistUpdateEvent(
+	ctx context.Context,
+	btcHeight uint64,
+	event *types.EventPowerDistUpdate,
+) {
+	store := k.powerDistUpdateEventStore(ctx, btcHeight)
+
+	// get event index
+	eventIdx := uint64(0) // event index starts from 0
+	iter := store.ReverseIterator(nil, nil)
```
KonradStaniec, line 239: maybe separate function `GetLastEventIdxAtheight` ?

KonradStaniec (approved):
only minor stuff:) great work!

x/btcstaking/keeper/btc_delegations.go
```diff
@@ -2,19 +2,160 @@ package keeper
...
 }
 
+// GetBTCDelegation gets the BTC delegation with a given staking tx hash
+func (k Keeper) GetBTCDelegation(ctx context.Context, stakingTxHashStr string) (*types.BTCDelegation, error) {
+	// decode staking tx hash string
+	stakingTxHash, err := chainhash.NewHashFromStr(stakingTxHashStr)
+	if err != nil {
+		return nil, err
+	}
+	btcDel := k.getBTCDelegation(ctx, *stakingTxHash)
```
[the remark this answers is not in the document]
SebastianElvis: Here `getBTCDelegation` is supposed to be an internal function whose caller might not necessarily treat a nil response as an error. External callers can use `GetBTCDelegation` which always returns an error if the BTC delegation is not found. Wdyt?

x/btcstaking/keeper/grpc_query.go
SebastianElvis: Probably not needed, since if the function hits here, then the BTC delegation index (corresponding to `value` at L219) must not be empty, thus there must be BTC delegations under this BTC delegator.

x/btcstaking/keeper/msg_server.go
SebastianElvis: Hmm if covenant emulators somehow delay a bit such that the BTC delegation's timelock has already expired, then the covenant signatures need to be rejected. This check is supposed to serve this scenario. Wdyt?

x/btcstaking/keeper/power_dist_change.go
SebastianElvis: So far this is the only place that retrieves all events at a certain index so I would prefer keeping it here. Wdyt?

x/btcstaking/keeper/power_dist_change.go
SebastianElvis: Same as above, this is the only place that retrieves the last event index. Lmk if you still prefer moving it to a separate function

SebastianElvis:
@KonradStaniec @gitferry Thanks for the comments. I have answered or addressed them under a new PR https://github.com/babylonchain/babylon/pull/523 to make the commit history of this PR clean (This PR probably needs to be merged via a merge commit). Please feel free to take a look

x/btcstaking/keeper/msg_server.go
gitferry: We don't add it as valid covenant sig but we dont need to consider it as an error, right?

x/btcstaking/keeper/btc_delegations.go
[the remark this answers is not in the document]
gitferry: yeah, maybe we need to add a contract here about nil check is needed

x/btcstaking/keeper/msg_server.go
SebastianElvis: I see, make sense :+1:

x/btcstaking/keeper/power_dist_change.go
KonradStaniec: I mean, for me this whole clousre just distracts attention form what this function is doing, also you need to add this comment to explain why it is here, with separate function the meaning will be clear and there will be no need for comments, so yea I would prefer separate function.

x/btcstaking/keeper/power_dist_change.go
KonradStaniec: again this whole ceremony:
```
	// get event index
	eventIdx := uint64(0) // event index starts from 0
	iter := store.ReverseIterator(nil, nil)
	defer iter.Close()
	if iter.Valid() {
		// if there exists events already, event index will be the subsequent one
		eventIdx = sdk.BigEndianToUint64(iter.Key()) + 1
	}
```
rises a lot of question, while if it were in separate function it would be clear what it does at first glance.

So I would prefer for it be separate function. (it can be private helper)

PR #463, opened 2024-02-09
btcstaking: KV stores recording finality provider / BTC delegation state updates

SebastianElvis:
Fixes [BM-1203](https://babylon-chain.atlassian.net/browse/BM-1203) (step 2 of [BM-1995](https://babylon-chain.atlassian.net/browse/BM-1195))

This PR 

- adds a KV store for events that affect voting power distribution. The key is `(BTC height || index)` and value is a new event. The event can be `EventSlashedFinalityProvider` indicating that a finality provider is slashed, or `EventBTCDelegationStateUpdate` indicating that a BTC delegation's state has updated.
- records such events upon corresponding scenarios.

This PR allows Babylon to track all state transitions of BTC delegations and finality providers, thereby enabling calculation of voting power distribution by reconciling prior one with these new events.

TODOs in subsequent PRs:
- [ ] new algorihtm for updating voting power distribution ([BM-1205](https://babylon-chain.atlassian.net/browse/BM-1205))


SebastianElvis:
Updated the PR according to an offline discussion with @KonradStaniec The updated PR now only introduces a single KV store. It serves as a queue of events that affect voting power distribution update. This simplifies the event processing and data model, yet preserving the order of events happening on chain. Please feel free to review.

x/btcstaking/keeper/msg_server.go
```diff
@@ -536,13 +546,21 @@ func (ms msgServer) AddCovenantSigs(goCtx context.Context, req *types.MsgAddCove
...
-	// notify subscriber
-	event := &types.EventBTCDelegationStateUpdate{
-		StakingTxHash: req.StakingTxHash,
-		NewState:      types.BTCDelegationStatus_ACTIVE,
-	}
-	if err := ctx.EventManager().EmitTypedEvent(event); err != nil {
-		panic(fmt.Errorf("failed to emit EventBTCDelegationStateUpdate for the new active BTC delegation: %w", err))
+	// If reaching the covenant quorum after this msg, the BTC delegation becomes
+	// active. Then, record and emit this event
+	if len(btcDel.CovenantSigs) == int(params.CovenantQuorum) {
```
gitferry, line 551: This check has been done in previous checks: Line 431

x/btcstaking/keeper/power_dist_change.go
```diff
@@ -0,0 +1,172 @@
...
+	types.RecordInactiveFinalityProviders(numFPs - numActiveFPs)
+	numStakedBTCs := float32(numStakedSats / SatoshisPerBTC)
+	types.RecordMetricsKeyStakedBitcoins(numStakedBTCs)
+	// record metrics for BTC delegations
+	for status, num := range numDelsMap {
+		types.RecordBTCDelegations(num, status)
+	}
+
+	// filter out top `MaxActiveFinalityProviders` active finality providers in terms of voting power
+	maxNumActiveFPs := k.GetParams(ctx).MaxActiveFinalityProviders
```
gitferry, line 79:
```suggestion
	paramMaxNumActiveFps := params.MaxActiveFinalityProviders
```
Also, this line can be inserted before Line 68

gitferry (commented):
Nice work! Some initial comments:

x/btcstaking/keeper/finality_providers.go
```diff
@@ -39,13 +38,18 @@ func (k Keeper) GetFinalityProvider(ctx context.Context, fpBTCPK []byte) (*types
 // SlashFinalityProvider slashes a finality provider with the given PK
 // A slashed finality provider will not have voting power
 func (k Keeper) SlashFinalityProvider(ctx context.Context, fpBTCPK []byte) error {
+	// ensure finality provider exists
 	fp, err := k.GetFinalityProvider(ctx, fpBTCPK)
 	if err != nil {
 		return err
```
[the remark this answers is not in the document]
SebastianElvis: Here `err` is going to be `types.ErrFpNotFound` if `GetFinalityProvider` returns error ([ref](https://github.com/babylonchain/babylon/blob/d89c2e44358bb0bd34099fd70745bc6c8a3c962c/x/btcstaking/keeper/finality_providers.go#L29-L31))

x/btcstaking/keeper/msg_server.go
gitferry: Ah, I see. Would it be clearer if we call `btcDel.HasCovenantQuorums` here?

gitferry (approved):
Nice work!

x/btcstaking/keeper/msg_server.go
SebastianElvis: `HasCovenantQuorums` returns `true` if `len(btcDel.CovenantSigs) > int(params.CovenantQuorum)` as well. Here we only want the event to be emitted once upon `len(btcDel.CovenantSigs) == int(params.CovenantQuorum)`.

x/btcstaking/keeper/msg_server.go
```diff
@@ -386,6 +386,16 @@ func (ms msgServer) CreateBTCDelegation(goCtx context.Context, req *types.MsgCre
...
 
+	// record event that the BTC delegation becomes pending at this height
+	pendingEvent := types.NewEventPowerDistUpdateWithBTCDel(event)
+	ms.addPowerDistUpdateEvent(ctx, btcTip.Height, pendingEvent)
+	// record event that the BTC delegation will become unbonded at endHeight-w
+	unbondedEvent := types.NewEventPowerDistUpdateWithBTCDel(&types.EventBTCDelegationStateUpdate{
+		StakingTxHash: stakingTxHash.String(),
+		NewState:      types.BTCDelegationStatus_UNBONDED,
+	})
+	ms.addPowerDistUpdateEvent(ctx, endHeight-wValue, unbondedEvent)
```
KonradStaniec, line 397: it would be great to have some table driven tests showing corner cases for all those values (K, W, Current_BTC_Tip, Staking_time) i.e showing that even when stking time is minimal like equal to k+w , we won't register this event in the past and won't save it before last indexed btc height. (then expiry will not be processed ever)

KonradStaniec (approved):
Great work! It would be nice to have some table tests for corner cases but this is not a blocker for now

x/btcstaking/keeper/msg_server.go
SebastianElvis: Hmm isn't this guaranteed since the timelock [must](https://github.com/babylonchain/babylon/blob/3757f10d5511044ab5897541d029a2df1ee78ee3/x/btcstaking/keeper/msg_server.go#L238-L241) be at least `w+1`? With `w+1` timelock the unbonded event will be recorded at tip height + 1 so the unbonded event cannot be registered in the past. If we allow `endHeight-wValue` to be smaller than tip height then that's a bug in `CreateBTCDelegation` message handler, which is already tested.

`k` seems less relevant here because it does not contribute to the timelock.

KonradStaniec: tbh I needed to draw myself a timeline to convince myself that everything is fine 😅 I think what made me worry is the fact that we calculate `endheight` as `stakingTxHeader.Height + uint64(req.StakingTime)` and staking `stakingTxHeader.Height` is already `k` deep and at first glance it looked suspicious.

Thats why I think table driven tests would be good to show that even for corner cases  we never register event in the past.  This is not critical for now, but I defiantly worry that in the future during some refactoring someone will fiddle with those numbers a we will missing it.

PR #462, opened 2024-02-09
fix: ensuring BTC delegation exists for certain handlers

SebastianElvis:

PR #464, opened 2024-02-09
Fix bug when delegation does not exists

KonradStaniec:
Git cleaned up version of @SebastianElvis fix - https://github.com/babylonchain/babylon/pull/462


gitferry (approved):
Nice find & fix!

PR #466, opened 2024-02-09
Unmarshalling perf

KonradStaniec:
Pr in spirit of - https://github.com/babylonchain/babylon/pull/441

Latest flamegraphs from load testing shown that, deserialzing btc transaction and signatures takes most time in code paths which retrieve a lot of delegations. This pr removes some unnecessary checks from db perspective.

This is not proper solution, but it can be some kind of stop-gap if we we would need around 20% perf boost (based on begin block benchmarks). It is also backward compatbile with current testnet and all the programs so this is easy gain.

### DIFF
--- a/types/btc_schnorr_sig.go
+++ b/types/btc_schnorr_sig.go
@@ -2,7 +2,6 @@ package types
 
 import (
 	"encoding/hex"
-	"errors"
 
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
 )
@@ -69,15 +68,6 @@ func (sig BIP340Signature) MarshalTo(data []byte) (int, error) {
 }
 
 func (sig *BIP340Signature) Unmarshal(data []byte) error {
-	newSig := BIP340Signature(data)
-
-	// ensure that the bytes can be transformed to a *schnorr.Signature object
-	// this includes all format checks
-	_, err := newSig.ToBTCSig()
-	if err != nil {
-		return errors.New("bytes cannot be converted to a *schnorr.Signature object")
-	}
-
 	*sig = data
 	return nil
 }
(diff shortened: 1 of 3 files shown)


types/btc_schnorr_sig.go
```diff
@@ -69,15 +68,6 @@ func (sig BIP340Signature) MarshalTo(data []byte) (int, error) {
...
-	newSig := BIP340Signature(data)
-
-	// ensure that the bytes can be transformed to a *schnorr.Signature object
-	// this includes all format checks
-	_, err := newSig.ToBTCSig()
-	if err != nil {
-		return errors.New("bytes cannot be converted to a *schnorr.Signature object")
-	}
-
 	*sig = data
```
SebastianElvis, line 71: Given that we don't perform verification here, `NewBIP340Signature` won't perform any verification either. Since `NewBIP340Signature` will be the function that takes inputs from the external users, we need to move the verification removed here to `NewBIP340Signature`.

x/btcstaking/types/msg.go
```diff
@@ -145,6 +168,13 @@ func (m *MsgAddCovenantSigs) ValidateBasic() error {
 	if m.UnbondingTxSig == nil {
 		return fmt.Errorf("empty covenant signature")
 	}
+
+	_, err := m.UnbondingTxSig.ToBTCSig()
+
+	if err != nil {
```
SebastianElvis, line 174:
```suggestion
	if _, err := m.UnbondingTxSig.ToBTCSig(); err != nil {
```
IIRC this is the practice recommended by Go. Same comments for other places in this file

SebastianElvis (approved):

PR #469, opened 2024-02-14
btcstaking: editing finality provider

SebastianElvis:
Fixes [BM-1254](https://babylon-chain.atlassian.net/browse/BM-1254)

This PR adds the functionality of editing an existing finality provider, including its description and commission. This is done via a new message `MsgEditFinalityProvider`, similar to `MsgEditValidator` in [Cosmos SDK](https://github.com/cosmos/cosmos-sdk/blob/33868606a86864dcc68510a3c7cc80c449ea5ec0/proto/cosmos/staking/v1beta1/tx.proto#L83-L102).


proto/babylon/btcstaking/v1/tx.proto
```diff
@@ -56,6 +58,26 @@ message MsgCreateFinalityProvider {
...
 message MsgCreateFinalityProviderResponse {}
 
+// MsgEditFinalityProvider is the message for editing an existing finality provider
+message MsgEditFinalityProvider {
+  option (cosmos.msg.v1.signer) = "signer";
+
+  // NOTE: this signer needs to correspond to babylon_pk of the finality provider
+  string signer = 1;
+  // btc_pk is the Bitcoin secp256k1 PK of the finality provider to be edited
+  bytes btc_pk = 2;
```
gitferry, line 68: Can we use `BIP340PubKey` format to be consistent with `MsgCreateFinalityProvider`

gitferry (approved):
Nicely done!

proto/babylon/btcstaking/v1/tx.proto
SebastianElvis: This is intentional -- since https://github.com/babylonchain/babylon/pull/438 we will refactor all APIs to separate DB objects and API objects. As a result `BIP340PubKey` will be a DB object, and shall not be used in messages. Other queries and txs will be revised in subsequent PRs (https://babylon-chain.atlassian.net/browse/BM-949)

PR #472, opened 2024-02-14
Update README.md

HarveyV:
Link to a broken user guide document is fixed. it was https://docs.babylonchain.io/docs/user-guides and now is https://docs.babylonchain.io/docs/user-guides/overview


SebastianElvis (approved):
Thank you!

PR #474, opened 2024-02-15
btcstaking: query finality provider

SebastianElvis:
This PR adds a query to finality provider with the given BTC PK.

Context: https://babylon-chain.slack.com/archives/C06CJ0J6UJ3/p1707884423920019?thread_ts=1707882635.267279&cid=C06CJ0J6UJ3

### DIFF
--- a/x/btcstaking/client/cli/query.go
+++ b/x/btcstaking/client/cli/query.go
@@ -23,12 +23,40 @@ func GetQueryCmd(queryRoute string) *cobra.Command {
 	}
 
 	cmd.AddCommand(CmdQueryParams())
+	cmd.AddCommand(CmdFinalityProvider())
 	cmd.AddCommand(CmdFinalityProviders())
 	cmd.AddCommand(CmdBTCDelegations())
 	cmd.AddCommand(CmdFinalityProvidersAtHeight())
 	cmd.AddCommand(CmdFinalityProviderPowerAtHeight())
 	cmd.AddCommand(CmdActivatedHeight())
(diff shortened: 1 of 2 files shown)


x/btcstaking/client/cli/query.go
```diff
@@ -33,6 +34,33 @@ func GetQueryCmd(queryRoute string) *cobra.Command {
...
+		Use:   "finality-provider [fp_btc_pk_hex]",
+		Short: "retrieve a finality provider",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			clientCtx := client.GetClientContextFromCmd(cmd)
+			queryClient := types.NewQueryClient(clientCtx)
+			res, err := queryClient.FinalityProvider(
+				cmd.Context(),
+				&types.QueryFinalityProviderRequest{
+					FpBtcPkHex: args[0],
```
gitferry, line 48: Do we need to verify the pk first?

gitferry (approved):
Lgtm!

x/btcstaking/client/cli/query.go
SebastianElvis: Hmm I would say it's not necessary. The node side will return error if the pk is malformed anyway. It's a query so does not have implication on gas. Wdyt?

gitferry: I see. Since we are passing the PK as string type, we need to check the type anyway. All good then

PR #475, opened 2024-02-15
hotfix: Select BlockHash that has achieved consensus

vitsalis:
Joint work with the core team. 

CometBFT does not verify vote extensions that are sent after the 2/3 threshold has been surpassed. This leads to blocks that can include invalid vote extensions.

The existing code makes the assumption that all vote extensions have been verified and are valid. Due to this, it chooses the first vote extension available to extract the block hash in order to build a checkpoint to include in the proposal. However, if the vote is invalid, the proposal constructed is invalid and rejected by everyone.

This led to a chain halt. What happened:
- At the end of the epoch, validators send vote extensions with their BLS signatures.
- Due to some validators not having set up their BLS keys properly, empty vote extensions were included in the block.
- The vote extensions were ordered in such a way, that the first one did not contain a valid block hash (an empty one), leading to an invalid checkpoint being included in the proposal.
- Everyone else seeing this proposal rejected the invalid checkpoint.

The cycle continued with proposers building invalid proposals and them getting rejected, leading to a chain halt.

Is it consensus breaking: Should not be unless there was a malicious vote extension that was not validated and used for the checkpoint. We are verifying that this did not happen on the testnet.

Open Questions:
- Why did some of the validators produce invalid vote extensions? Is it due to an invalid setup? If so, how can we disallow people running validators with an invalid setup (e.g. throw error).
- Why did the vote extensions that we found were empty? Even if a vote extension is not validated after consensus is reached, a validator should not send an empty vote extension.


x/checkpointing/proposal.go
```diff
@@ -180,15 +181,51 @@ func (h *ProposalHandler) getValidBlsSigs(ctx sdk.Context, extendedVotes []abci.
...
+			maxPower = power
+		}
+		totalPower += power
+	}
+	if len(resBlockHash) == 0 {
+		return nil, fmt.Errorf("could not find the block hash")
+	}
+	// Verify that the voting power committed to the found block hash is
+	// more than 2/3 of the total voting power.
+	if maxPower*3 < totalPower*2 {
```
KonradStaniec, line 221: cosmos sdk `ValidateVoteExtensions` uses something like:
```
	// If the sum of the voting power has not reached (2/3 + 1) we need to error.
	if requiredVP := ((totalVP * 2) / 3) + 1; sumVP < requiredVP {
		return fmt.Errorf(
			"insufficient cumulative voting power received to verify vote extensions; got: %d, expected: >=%d",
			sumVP, requiredVP,
		)
	}
```

for validating cumulative voting power maybe we can use the same approach

gitferry, line 221:
```suggestion
	if maxPower*3 <= totalPower*2 {
```
The accumulated power should be > 2/3 total power, right?

gitferry (approved):

KonradStaniec (approved):

RafilxTenfen (approved):

fishermanymc:
Great team work!

One small note triggered by the following line

> The existing code makes the assumption that all vote extensions have been verified and are valid.

An honest node should always verify all the data without trusting 3rd parties. In particular, it should not assume that 2/3 of the validators are honest.

PR #476, opened 2024-02-15
chore: collect most voted ext while validating vote extension 

RafilxTenfen:


RafilxTenfen:
related https://github.com/babylonchain/babylon/pull/475

PR #482, opened 2024-02-16
Setup a test suite for testing proposals

KonradStaniec:
Test suite implementation with initial test cases. I am still unsure wheter it is best to use mocks or real keeper, we can re-consider this after adding more test cases.


x/checkpointing/proposal_test.go
```diff
@@ -0,0 +1,196 @@
...
+	for _, info := range infos {
+		out = append(out, info.Info)
+	}
+	return out
+}
+
+func SetupProposalHandler(
+	t *testing.T,
+	c *gomock.Controller) (*checkpointing.ProposalHandler, *mocks.MockCheckpointingKeeper, sdk.Context) {
+	ek := mocks.NewMockCheckpointingKeeper(c)
```
SebastianElvis, line 44: Given that the [test helper](https://github.com/babylonchain/babylon/tree/dev/testutil/helper) already supports testing ABCI++ to some extent, does it make sense to reuse the test helper to build such a test suite? The mocks make extra assumptions on the system and building a new test suite introduces some redundancy

KonradStaniec: So those tests and tests provided by test helper, serve two different purposes.

Tests provided by helper, shows how our domain logic fits within cosmos-sdk framework. It test our domain logic a bit, but there is a lot of other stuff (panic catches, different states, some databases). In my mind this sits at the level integration tests.

Main purposes of this tests is show to external callers what is expected behaviour of our module public api in different set of scenarios.( and of course test those scenarios, but in my mind this a bit of by product)
Having such test suite serves both, as testing for our domain logic and also documentation of behaviour for any other modules. In case of problem like, `oh I wonder, what happen in case of all empty vote extensions` the person, doesn't need to even consult source code , but can look at test cases at see the expected behaviour is to just reject proposal.

Even people which will modify the code in the future will see what is expected set of behaviours given module should honor.

testutil/mocks/proposal_keeper.go
SebastianElvis, line 1: Shouldn't this file be somewhere under `x/checkpointing` ?

x/checkpointing/proposal_test.go
```diff
@@ -0,0 +1,479 @@
...
+	valAddress, err := sdk.ValAddressFromBech32(v.Keys.ValidatorAddress)
+	require.NoError(t, err)
+	return valAddress
+}
+
+func (v *TestValidator) BlsPubKey() bls12381.PublicKey {
+	return *v.Keys.BlsKey.Pubkey
+}
+
+func genNTestValidators(t *testing.T, n int) []TestValidator {
```
SebastianElvis, line 108: Probably we can have a test suite that wraps a list of validators, ctx, and the checkpointing keeper?

x/checkpointing/proposal_test.go
```diff
@@ -0,0 +1,479 @@
...
+		Txs:        [][]byte{},
+		LocalLastCommit: cbftt.ExtendedCommitInfo{
+			Round: 0,
+			Votes: votes,
+		},
+		Height: height,
+	}
+}
+
+func randomBlockHash() checkpointingtypes.BlockHash {
```
SebastianElvis, line 193: This can go to `datagen/`

SebastianElvis (commented):
Great work! However there might be some duplications with existing tests based on `testutil/helper` and we might be able to simplify the code by using `github.com/stretchr/testify/suite`. Wdyt?

testutil/mocks/proposal_keeper.go
KonradStaniec: tbh I am not sure, there is a lot of mocks under `"github.com/babylonchain/babylon/testutil/mocks"` so it seemed like a good place to add another one.

x/checkpointing/proposal_test.go
KonradStaniec: sure, there is already `GenRandomBlockHash` there which I have missed 👍

x/checkpointing/proposal_test.go
KonradStaniec: not sure this will help much as for each scenario list of validators is potentially different and there are potentialy different expectations on mock keep. I will refactor a bit scenario setup function, but if we want to cover a lot of different cases then there will be some inherent complexity in those tests.

x/btcstaking/client/cli/query.go
```diff
@@ -61,6 +62,35 @@ func CmdFinalityProvider() *cobra.Command {
 	return cmd
 }
 
+func CmdDelegation() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "delegatopn [staking_tx_hash_hex]",
```
SebastianElvis, line 67:
```suggestion
		Use:   "delegation [staking_tx_hash_hex]",
```

SebastianElvis (approved):
Would be great to get another pass from @gitferry who has also written some tests on this :+1:

x/btcstaking/client/cli/query.go
KonradStaniec: ahh this wasn't supposed to be part of pr, but on the other hand this command comes in handy for some delegation debugging, so it is probably good to have it 😅

x/checkpointing/proposal_test.go
```diff
@@ -0,0 +1,469 @@
...
+					validator := val
+					ek.EXPECT().GetPubKeyByConsAddr(gomock.Any(), sdk.ConsAddress(validator.ValidatorAddress(t).Bytes())).Return(validator.ProtoPubkey(), nil).AnyTimes()
+
+					if i < invalidValidBlsSig {
+						ek.EXPECT().VerifyBLSSig(gomock.Any(), validatorAndExtensions.Extensions[i].ToBLSSig()).Return(checkpointingtypes.ErrInvalidBlsSignature).AnyTimes()
+					} else {
+						ek.EXPECT().VerifyBLSSig(gomock.Any(), validatorAndExtensions.Extensions[i].ToBLSSig()).Return(nil).AnyTimes()
+					}
+					ek.EXPECT().GetBlsPubKey(gomock.Any(), validator.ValidatorAddress(t)).Return(validator.BlsPubKey(), nil).AnyTimes()
+					// empty vote extension
```
gitferry, line 353: leftover?

x/checkpointing/proposal_test.go
```diff
@@ -0,0 +1,469 @@
...
+				return &Scenario{
+					TotalPower:   totalPower,
+					ValidatorSet: validatorAndExtensions.Vals,
+					Extensions:   signedVoteExtensions,
+				}
+			},
+			expectError: true,
+		},
+		{
+			name: "1/3 of validators provided invalid bls signature",
```
gitferry, line 333: maybe we can add one more case where <1/3 of validators provided invalid BLS signature and the expected error is false

gitferry (approved):
Awesome work! I like the idea of abstracting the checkpoint keeper

PR #487, opened 2024-02-20
chore: create client friendly response for BtcCheckpointInfo query

RafilxTenfen:
- At `QueryBtcCheckpointInfo` return `BTCCheckpointInfoResponse` instead of `BTCCheckpointInfo`

### DIFF
--- a/proto/babylon/btccheckpoint/v1/query.proto
+++ b/proto/babylon/btccheckpoint/v1/query.proto
@@ -2,6 +2,7 @@ syntax = "proto3";
 package babylon.btccheckpoint.v1;
 
 import "gogoproto/gogo.proto";
+import "cosmos_proto/cosmos.proto";
 import "google/api/annotations.proto";
 import "cosmos/base/query/v1beta1/pagination.proto";
 import "babylon/btccheckpoint/v1/params.proto";
@@ -55,7 +56,7 @@ message QueryBtcCheckpointInfoRequest {
 
 // QueryBtcCheckpointInfoResponse is response type for the
 // Query/BtcCheckpointInfo RPC method
-message QueryBtcCheckpointInfoResponse { BTCCheckpointInfo info = 1; }
+message QueryBtcCheckpointInfoResponse { BTCCheckpointInfoResponse info = 1; }
 
 // QueryBtcCheckpointsInfoRequest is request type for the
 // Query/BtcCheckpointsInfo RPC method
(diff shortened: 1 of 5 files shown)


x/btccheckpoint/types/query.go
```diff
@@ -0,0 +1,52 @@
...
+		Transaction: hex.EncodeToString(ti.Transaction),
+		Proof:       hex.EncodeToString(ti.Proof),
+	}
+}
+
+// ToResponse parses a TransactionKeyResponse into a query response tx key struct.
+func (tk *TransactionKey) ToResponse() *TransactionKeyResponse {
+	return &TransactionKeyResponse{
+		Index: tk.Index,
+		Hash:  hex.EncodeToString(*tk.Hash),
```
KonradStaniec, line 22: So all btc hash responses need to be marshalled to hex in special bitcoin way (for some reason marshall hashes to hex in reverse order), here you can use `tk.Hash.MarshallHex()`. Another option would be to leave hash as custom hash type.

x/btccheckpoint/types/query.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+	bestSubVigAddrs := make([]*CheckpointAddressesResponse, len(b.BestSubmissionVigilanteAddressList))
+	for i, addrs := range b.BestSubmissionVigilanteAddressList {
+		bestSubVigAddrs[i] = addrs.ToResponse()
+	}
+
+	return &BTCCheckpointInfoResponse{
+		EpochNumber:                        b.EpochNumber,
+		BestSubmissionBtcBlockHeight:       b.BestSubmissionBtcBlockHeight,
+		BestSubmissionBtcBlockHash:         hex.EncodeToString(*b.BestSubmissionBtcBlockHash),
```
KonradStaniec, line 48: bitcooin hex encoding - `b.BestSubmissionBtcBlockHash.MarshallHex()`

proto/babylon/btccheckpoint/v1/query.proto
```diff
@@ -89,3 +90,58 @@ message QueryEpochSubmissionsResponse {
...
+// - the full tx content
+// - the Merkle proof that this tx is on the above position
+message TransactionInfoResponse {
+  // key is the position (txIdx, blockHash) of this tx on BTC blockchain
+  // Although it is already a part of SubmissionKey, we store it here again
+  // to make TransactionInfo self-contained.
+  // For example, storing the key allows TransactionInfo to not relay on
+  // the fact that TransactionInfo will be ordered in the same order as
+  // TransactionKeys in SubmissionKey.
+  TransactionKeyResponse key = 1;
```
KonradStaniec, line 122: I think for external consumption is more natural to have:
```
message TransactionInfoResponse {
  // Index Bitcoin Transaction index in block.
  uint32 index = 1;
  // Hash BTC Header hash as hex.
  string hash = 2;
  // transaction is the full transaction data as str hex.
  string transaction = 3;
  // proof is the Merkle proof that this tx is included in the position in `key`
  string proof = 4;
}
```

Then you have all data describing transaction in one message.

KonradStaniec (approved):
lgtm! 

I have created new label `API BREAKING` so have easy way to finds pr-s which will break api. Lets use it subsequent pr-s.

PR #490, opened 2024-02-20
Update btcstaking_test.go

leonalAbel:
Error handling made independent of require function


SebastianElvis:
Thanks, but it's not clear about the problem of doing `require.Error` here.

PR #492, opened 2024-02-22
fix: Disable sending of empty vote extension

gitferry:
This PR fixed the issue where empty vote extension will be sent when some critical error happens during `ExtendVote`. We should not allow empty vote extension to be sent as we cannot ensure all the vote extensions will be checked by VerifyVoteExtension due to the [issue](https://github.com/cometbft/cometbft/issues/2361).

This PR used panics to replace returning errors because CometBFT will [still send empty vote extension](https://github.com/cosmos/cosmos-sdk/blob/7dbed2fc0c3ed7c285645e21cb1037d8810372ae/baseapp/abci.go#L612) upon errors from `ExtendVote`.

This PR added 3 fuzz tests to cover all the cases where panic could happen, specifically when:
* the BLS private key is missing
* the signing block hash has invalid format
* the signer is not in the validator set


SebastianElvis:
Also is it possible to panick even ealier if BLS is not correctly configured, e.g., when starting the node?

gitferry:
> Also is it possible to panick even ealier if BLS is not correctly configured, e.g., when starting the node?

I think we have the check here: https://github.com/babylonchain/babylon/blob/c472fe407498f78ff644fc1545cc450bac51ea54/cmd/babylond/cmd/root.go#L276

x/checkpointing/vote_ext.go
```diff
@@ -49,22 +57,25 @@ func (h *VoteExtensionHandler) ExtendVote() sdk.ExtendVoteHandler {
...
 		_, _, err := curValSet.FindValidatorWithIndex(signer)
 		if err != nil {
-			// not being a validator is not an error
-			h.logger.Info("the BLS signer is not in the validator set", "signer_address", signer.String())
-			return emptyRes, nil
+			// NOTE: this indicates programmatic error because ExtendVote
+			// should not be invoked if the validator is not in the
+			// active set according to:
+			// https://github.com/cometbft/cometbft/blob/a17290f6905ef714761f12c1f82409b0731e3838/consensus/state.go#L2434
+			panic(fmt.Errorf("the BLS signer %s is not in the validator set", signer.String()))
```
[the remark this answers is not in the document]
gitferry: Technically yes, but we still can run into this case if (1) CometBFT has a bug, or (2) our implementation of the checkpointing module has a bug that stores/fetches wrong validator set.

Sending `emptyRes` is good in the sense that it does not block consensus. However, due to the [issue](https://github.com/cometbft/cometbft/issues/2361) CometBFT, not all the vote extensions will be verified by `VerifyVoteExtension`. Therefore, we should disallow the sending of empty vote extensions in any case.

x/checkpointing/vote_ext.go
```diff
@@ -49,22 +57,25 @@ func (h *VoteExtensionHandler) ExtendVote() sdk.ExtendVoteHandler {
...
 		}
 
 		// 2. sign BLS signature
 		blsSig, err := k.SignBLS(epoch.EpochNumber, req.Hash)
 		if err != nil {
-			// the returned error will lead to panic
-			return emptyRes, fmt.Errorf("failed to sign BLS signature at epoch %v, height %v",
-				epoch.EpochNumber, req.Height)
+			// NOTE: this indicates misconfiguration of the BLS key
+			panic(fmt.Errorf("failed to sign BLS signature at epoch %v, height %v",
```
KonradStaniec, line 71: General comment: 

I am not sure those panics accomplish our desired goal as every panic will be catched by cosmos sdk - https://github.com/cosmos/cosmos-sdk/blob/release/v0.50.x/baseapp/abci.go#L597, and transformed into empty vote extension.

It can happen for any panic in abci handler, this means those panics can be lost in sea of other panics like out of gas panics.

Maybe alternative would be to create `ErrInvalidValidatorConfiguration` with some super descriptive message that will be more visible to potential validators and return it in those cases ? 

It will help also with testiblity as then in testsuire for vote extensions, you can write cases for this scenario and assert that returned error is really `ErrInvalidValidatorConfiguration` 

Imo the only panic which would help with this situation is the panic during node startup, which say `I failed to properly configure your validator: this is the error: err`

gitferry: > I am not sure those panics accomplish our desired goal as every panic will be catched by cosmos sdk - https://github.com/cosmos/cosmos-sdk/blob/release/v0.50.x/baseapp/abci.go#L597, and transformed into empty vote extension.

These panics actually will be transformed into a [unified error](https://github.com/cosmos/cosmos-sdk/blob/fcfc09995373ec5fb9b9f6a5fe7c82edb016f23e/baseapp/abci.go#L605) and caught by CometBFT, leading to a [panic](https://github.com/cometbft/cometbft/blob/a17290f6905ef714761f12c1f82409b0731e3838/consensus/state.go#L2388).

And I think with the panics, we can see from the log what reason has led to the panic. If we don't panic but return errors (even a descriptive one), the errors will be hidden as the node will be running fine

KonradStaniec: oh you are right.

Just verified that panic in extend vote are actually transformed into:
```
2:02PM INF received complete proposal block hash=0F1AB6D35BE80CEEBF2B6273D55DE9AB72C27D4996A50532DE9A3E2B47774D57 height=1 module=consensus
2:02PM ERR panic recovered in ExtendVote hash=0F1AB6D35BE80CEEBF2B6273D55DE9AB72C27D4996A50532DE9A3E2B47774D57 height=1 module=server panic=null
2:02PM ERR CONSENSUS FAILURE!!! err="ExtendVote call failed: recovered application panic in ExtendVote: wooot" module=consensus stack="goroutine 408 [running]:\nruntime/debug.Stack()\n\truntime/debug/stack.go:24 +0x64\ngithub.com/cometbft/cometbft/consensus.(*State).receiveRoutine.func2()\n\tgithub.com/cometbft/cometbft@v0.38.5/consensus/state.go:800 +0x44\npanic({0x103d5ba80?, 0x1400147ef60?})\n\truntime/panic.go:914 +0x218\ngithub.com/cometbft/cometbft/state.(*BlockExecutor).ExtendVote(_, {_, _}, _, _, {{{0xb, 0x0}, {0x1400256df0a, 0x6}}, {0x1400256df20, ...}, ...})\n\tgithub.com/cometbft/cometbft@v0.38.5/state/execution.go:340 +0x3e0\ngithub.com/cometbft/cometbft/consensus.(*State).signVote(0x14001d12700, 0x2, {0x140013cb940, 0x20, 0x20}, {0x165f39c?, {0x14000a789a0?, 0x140035d72ac?, 0x30?}}, 0x1?)\n\tgithub.com/cometbft/cometbft@v0.38.5/consensus/state.go:2378 +0x3d8\ngithub.com/cometbft/cometbft/consensus.(*State).signAddVote(0x14001d12700, 0x1?, {0x140013cb940, 0x20, 0x20}, {0xa?, {0x14000a789a0?, 0x0?, 0x0?}}, 0x0?)\n\tgithub.com/cometbft/cometbft@v0.38.5/consensus/state.go:2439 +0x1a4\ngithub.com/cometbft/cometbft/consensus.(*State).enterPrecommit(0x14001d12700, 0x1, 0x0)\n\tgithub.com/cometbft/cometbft@v0.38.5/consensus/state.go:1536 +0x1090\ngithub.com/cometbft/cometbft/consensus.(*State).addVote(0x14001d12700, 0x14001cd7930, {0x0, 0x0})\n\tgithub.com/cometbft/cometbft@v0.38.5/consensus/state.go:2296 +0x1570\ngithub.com/cometbft/cometbft/consensus.(*State).tryAddVote(0x14001d12700, 0x14001cd7930, {0x0?, 0x100e7e2b0?})\n\tgithub.com/cometbft/cometbft@v0.38.5/consensus/state.go:2056 +0x28\ngithub.com/cometbft/cometbft/consensus.(*State).handleMsg(0x14001d12700, {{0x10425d320, 0x140018981f8}, {0x0, 0x0}})\n\tgithub.com/cometbft/cometbft@v0.38.5/consensus/state.go:928 +0x32c\ngithub.com/cometbft/cometbft/consensus.(*State).receiveRoutine(0x14001d12700, 0x0)\n\tgithub.com/cometbft/cometbft@v0.38.5/consensus/state.go:855 +0x348\ncreated by github.com/cometbft/cometbft/consensus.(*State).OnStart in goroutine 1\n\tgithub.com/cometbft/cometbft@v0.38.5/consensus/state.go:397 +0xf0\n"
```

and not extension is being send.

KonradStaniec (approved):

PR #502, opened 2024-02-26
btcstaking: more tests on finality provider set rotation

SebastianElvis:
Fixes [BM-1160](https://babylon-chain.atlassian.net/browse/BM-1160)

This PR introduces more fuzz tests on finality provider set rotation, including

- voting power update during the full lifecyle of a slashed finality provider with BTC delegations
- voting power update during the full lifecyle of an expired BTC delegation
- active finality provider set rotation in the presence of changing BTC delegations

This is the final PR to the https://github.com/babylonchain/babylon/tree/base/index-optimise-btcstaking branch, finalising the development of this feature.

### DIFF
--- a/x/btcstaking/keeper/power_dist_change_test.go
+++ b/x/btcstaking/keeper/power_dist_change_test.go
@@ -80,38 +80,20 @@ func FuzzBTCDelegationEvents(f *testing.F) {
 		)
 
 		/*
-			at this point, there should be
-			- 1 event that BTC delegation becomes pending at current BTC tip
-			- 1 event that BTC delegation will become expired at end height - w
+			at this point, there should be 1 event that BTC delegation will become expired at end height - w
 		*/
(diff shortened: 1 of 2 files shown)


KonradStaniec (approved):
looks good! 

Another nice set of test would be to have tests in which:
- we generate some amount of events in database
- run begin block
-  check that voting power distriubution matches expectations

SebastianElvis:
I see your point. In fact this is already covered by `FuzzVotingPowerTable_ActiveFinalityProviders` and `FuzzVotingPowerTable_ActiveFinalityProviderRotation` in this PR where we do a bunch of delegations with different patterns, then execute `BeginBlock`, and finally assert consistency of voting power table. 

I'm not sure if it's a good idea to generate events directly in the tests since this requires making functions for generating events public. It's better to be kept private and not reachable from outside the keeper. Wdyt?

PR #514, opened 2024-02-29
Apply typo fixed and remake proto-gen

miles-six:


miles-six:
I will delate previous pr when this pr is mergerd,is it ok?

PR #531, opened 2024-03-06
Hot fix perf

KonradStaniec:
Fixes:
- unmarshalling performance
- improved iteration over delegations

### DIFF
--- a/x/btcstaking/keeper/keeper.go
+++ b/x/btcstaking/keeper/keeper.go
@@ -9,6 +9,7 @@ import (
 	"cosmossdk.io/log"
 	"github.com/babylonchain/babylon/x/btcstaking/types"
 	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/cosmos/cosmos-sdk/codec"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
@@ -60,7 +61,6 @@ func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 // the reward distribution cache used for distributing rewards once the block
 // is finalised by finality providers.
 func (k Keeper) BeginBlocker(ctx context.Context) error {
-	// index BTC height at the current height
 	k.IndexBTCHeight(ctx)
 
 	covenantQuorum := k.GetParams(ctx).CovenantQuorum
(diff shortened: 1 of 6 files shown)


SebastianElvis (approved):

x/btcstaking/keeper/keeper.go
```diff
@@ -70,24 +70,36 @@ func (k Keeper) BeginBlocker(ctx context.Context) error {
...
-	// prepare for recording finality providers with positive voting power
+	distInfos := make(map[chainhash.Hash]*types.BTCDelDistInfo)
+
+	k.IterateBTCDelsKeys(ctx, func(key *chainhash.Hash, btcDel *types.BTCDelegation) bool {
+		distInfo := &types.BTCDelDistInfo{
+			BabylonPk:   btcDel.BabylonPk,
+			VotingPower: btcDel.VotingPower(btcTipHeight, wValue, covenantQuorum),
+		}
+		if distInfo.VotingPower > 0 {
+			distInfos[*key] = distInfo
```
vitsalis, line 81: Let's do a `nil` check on `key`?

x/btcstaking/keeper/keeper.go
```diff
@@ -70,24 +70,36 @@ func (k Keeper) BeginBlocker(ctx context.Context) error {
...
 		func(fp *types.FinalityProvider) bool {
 			fpDistInfo := types.NewFinalityProviderDistInfo(fp)
 
-			// iterate over all BTC delegations under the finality provider
-			// in order to accumulate voting power and reward dist info for it
-			k.IterateBTCDelegations(ctx, fp.BtcPk, func(btcDel *types.BTCDelegation) bool {
-				// accumulate voting power and reward distribution cache
-				fpDistInfo.AddBTCDel(btcDel, btcTipHeight, wValue, covenantQuorum)
+			k.IterateBTCDelegationsHashes(ctx, fp.BtcPk, func(hash *chainhash.Hash) bool {
+				distInfo, found := distInfos[*hash]
```
vitsalis, line 96: Same here.

vitsalis (approved):

x/btcstaking/keeper/keeper.go
KonradStaniec: given that we control both implementation and we are making copy either way, it will be cleaner to not use pointer here at all but rather copy of the value

PR #533, opened 2024-03-06
chore: update query `EpochMsgs` to return `QueuedMessageResponse` 

RafilxTenfen:
- Add new struct `QueuedMessageResponse`
- Update query `EpochMsgs` to return `QueuedMessageResponse` instead of `QueuedMessage`


proto/babylon/epoching/v1/query.proto
```diff
@@ -108,7 +108,7 @@ message QueryCurrentEpochRequest {}
 message QueryCurrentEpochResponse {
   // current_epoch is the current epoch number
   uint64 current_epoch = 1;
-  // epoch_boundary is the height of this epoch's last block
+  // epoch_boundary is the height of this epoch's last blockf
```
SebastianElvis, line 111:
```suggestion
  // epoch_boundary is the height of this epoch's last block
```

SebastianElvis (approved):

proto/babylon/epoching/v1/query.proto
RafilxTenfen: done at https://github.com/babylonchain/babylon/pull/533/commits/3c3c5d4dfd1ac5e72c6087eeff6b9dd5e46fd068

PR #534, opened 2024-03-06
chore: update query `LatestEpochMsgs` to return `QueuedMessageResponse`

RafilxTenfen:
- Modified struct `QueuedMessageList` to use `QueuedMessageResponse` instead of `QueuedMessage`


x/epoching/types/query.go
```diff
@@ -25,3 +27,12 @@ func (q *QueuedMessage) ToResponse() *QueuedMessageResponse {
 		Msg:         q.UnwrapToSdkMsg().String(),
 	}
 }
+
+// NewQueuedMessagesResponse parses all the queued messages as response.
+func NewQueuedMessagesResponse(msgs []*QueuedMessage) (resp []*QueuedMessageResponse) {
```
SebastianElvis, line 32:
```suggestion
func NewQueuedMessagesResponse(msgs []*QueuedMessage) []*QueuedMessageResponse {
```

SebastianElvis (approved):

x/epoching/types/query.go
RafilxTenfen: done at https://github.com/babylonchain/babylon/pull/534/commits/df4982f734e281256043f1b98e831c5b84bf9f51

PR #536, opened 2024-03-07
chore: update query `ValidatorLifecycle` to return `ValidatorLifecycleResponse`

RafilxTenfen:
- Add new structs `ValidatorLifecycleResponse` and `ValStateUpdateResponse`
- Update query `ValidatorLifecycle` to return `ValidatorLifecycleResponse`


proto/babylon/epoching/v1/query.proto
```diff
@@ -158,7 +158,7 @@ message QueryValidatorLifecycleRequest { string val_addr = 1; }
 
 // QueryValidatorLifecycleResponse is the response type for the
 // Query/ValidatorLifecycle RPC method
-message QueryValidatorLifecycleResponse { ValidatorLifecycle val_life = 1; }
+message QueryValidatorLifecycleResponse { ValidatorLifecycleResponse val_life = 1; }
```
SebastianElvis, line 161: How about we simply define
```suggestion
message QueryValidatorLifecycleResponse { 
    string val_addr = 1;
    repeated ValStateUpdateResponse val_life = 2;
}
```
and get rid of `ValidatorLifecycleResponse`?

proto/babylon/epoching/v1/query.proto
```diff
@@ -232,3 +232,19 @@ message QueuedMessageList {
   uint64 epoch_number = 1;
   repeated QueuedMessageResponse msgs = 2;
 }
+
+// ValStateUpdateResponse is a message response that records a state update of a validator.
+message ValStateUpdateResponse {
+  BondState state = 1;
+  // StateDesc defines the descriptive state.
+  string state_desc = 2;
```
gitferry, line 240:
```suggestion
  // state_desc defines the descriptive state.
  string state_desc = 2;
```
A field of BondState is probably not needed as we already have `state_desc`

SebastianElvis (approved):

gitferry (approved):
Thanks for the fix!

PR #551, opened 2024-03-12
hotfix: Fix voting power query returns `0` regardless errors

gitferry:
Currently, `GetVotingPower` returns `0` without distinguishing whether there's a record in the voting power store or the fp indeed has `0` voting power.

This PR fixed the issue by adding an existence check of the voting power store within `GetVotingPower` and returning an error if it does not exist.


SebastianElvis (approved):
Nice catch!

KonradStaniec (approved):

Issue #564, opened 2024-03-13
Create new parameter `MinUnbondingOutputFraction` 


KonradStaniec:
Create new parameter which would require that unbonding output must be at lest `X%` of staking output (`X` should be in range of 80-90%, depending on other params). 

This is to avoid leaking value through fees, which would go to the miners. This makes system attackable by BTC miners.

PR #576, opened 2024-03-15
Min unbonding fraction

KonradStaniec:
fixes - https://github.com/babylonchain/babylon/issues/564

Adds new parameter which sets expectation in regard the value locked in unbonding output. This is necessary so that somobody controlling enogh hash power would not leak value from the system using fees.


x/btcstaking/types/params.go
```diff
@@ -64,6 +64,8 @@ func DefaultParams() Params {
 		// The default minimum unbonding time is 0, which effectively defaults to checkpoint
 		// finalization timeout.
 		MinUnbondingTime: 0,
+		// By default unbonding value is 0.8
+		MinUnbondingRate: sdkmath.LegacyNewDecWithPrec(8, 1), // 8 * 10^{-1} = 0.8
```
SebastianElvis, line 68: How do you compare this with implementing a `MaxUnbondingTxFeeSat` parameter? To me the latter is better as
- The root cause of the issue is the tx fee is leaked to the Bitcoin miners, so `MaxUnbondingTxFeeSat` seems a more straightforward fix.
- We have `MinSlashingTxFeeSat` (and in the future utility functions for calculating fees) so it seems easier for us.
- `MaxUnbondingTxFeeSat` is universal to all staking txs, i.e., independent to the staking tx output value.

KonradStaniec: Tbh I do not like either this or `MaxUnbondingTxFeeSat`, as we have too many parameters  as most probably it is easy to create set of parameters which does not make sense. 

Going over arguments:

>The root cause of the issue is the tx fee is leaked to the Bitcoin miners, so MaxUnbondingTxFeeSat seems a more straightforward fix

Imo both fixes address issue in pretty simple manner

> We have MinSlashingTxFeeSat (and in the future utility functions for calculating fees) so it seems easier for us.

We also have slashing rate, and btc have utilities to calculate rates,so easy is not factor here

> MaxUnbondingTxFeeSat is universal to all staking txs, i.e. independent to the staking tx output value.
 
This is good point. With `MaxUnbondingTxFeeSat` param, everybody is treated equally while `MinUnbondingRate` favours large stakers as they have more btc in staking output.

My pain point with `MaxUnbondingTxFeeSat` is that it implicitly influences staking value and currently we do not have way of setting minimal staking value  i.e lets say we will set it to `1000 sats/vb` (to give people a lot of leeway in setting fee as they may want to unbond quickly even in super high fee enviroments), unbonding tx with all signatures, will have ~200vb, this means that we will need to have `MaxUnbondingTxFeeSat  = 200000 sats` . This means that we should not accept staking transactions which has less than `400000` sats (to avoid leaking over 50% by fees in cae of miners attack)

So maybe tldr:
1. With current set of parameters, I prefer `rate` as it more easily express what we want to achieve without introducing additional problems
2. In world in which we have `MinStakingValue` (or even required `StakingValue` ) I prefer `MaxUnbondingTxFeeSat` , as it more close to the truth. 

My plan forward would be:
1. Merge this pr to not have vulnerability in our codebase 
2. There will be probably a lot of discussion for different parameters for phase-1 mainnet.
3. After discussions/and testnet, create ADR consolidating all the parameters (and explaining them in one place)
4. Implement ADR in our codebase

SebastianElvis: > My pain point with `MaxUnbondingTxFeeSat` is that it implicitly influences staking value and currently we do not have way of setting minimal staking value i.e lets say we will set it to `1000 sats/vb` (to give people a lot of leeway in setting fee as they may want to unbond quickly even in super high fee enviroments), unbonding tx with all signatures, will have ~200vb, this means that we will need to have `MaxUnbondingTxFeeSat = 200000 sats` . This means that we should not accept staking transactions which has less than `400000` sats (to avoid leaking over 50% by fees in cae of miners attack)

The `MinUnbondingRate` has a similar issue there -- one should provide a large enough staking tx output such that the unbonding tx can ever be accepted by Bitcoin miners. This implicitly imposes a minimal value for the staking value.
 
> 
> My plan forward would be:
> 
> 1. Merge this pr to not have vulnerability in our codebase
> 2. There will be probably a lot of discussion for different parameters for phase-1 mainnet.
> 3. After discussions/and testnet, create ADR consolidating all the parameters (and explaining them in one place)
> 4. Implement ADR in our codebase

Agreed, let's eliminate bugs first and then trim these parameters. Tbh it's a bit ichy to me that btcstaking module has so many parameters. At some point it would be good to revisit them a bit.

x/btcstaking/keeper/msg_server.go
```diff
@@ -401,6 +401,15 @@ func (ms msgServer) CreateBTCDelegation(goCtx context.Context, req *types.MsgCre
...
 	}
 
+	// this conversions must always succeed, as it is part of our params
+	minUnbondingRate := params.MinUnbondingRate.MustFloat64()
+	// Caluclate min unbonding output value based on staking output, use btc native multiplication
+	minUnbondingOutputValue := btcutil.Amount(stakingMsgTx.TxOut[newBTCDel.StakingOutputIdx].Value).MulF64(minUnbondingRate)
+
+	if btcutil.Amount(unbondingMsgTx.TxOut[0].Value) < minUnbondingOutputValue {
+		return nil, types.ErrInvalidUnbondingTx.Wrapf("unbonding output value must be at least %s, based on staking output", minUnbondingOutputValue)
+	}
```
SebastianElvis, line 411: The math operation here is a bit intensive so could we wrap them into a function?

SebastianElvis (approved):
Lgtm! This is a param change so @filippos47

x/btcstaking/types/params.go
KonradStaniec: >The MinUnbondingRate has a similar issue there -- one should provide a large enough staking tx output such that the unbonding tx can ever be accepted by Bitcoin miners. This implicitly imposes a minimal value for the staking value.

yup, thats true, although this does not break Babylon system, but rather makes things super inconvenient for staker i.e only staker is hurt if there is to little value in initial staking output

>Agreed, let's eliminate bugs first and then trim these parameters. Tbh it's a bit ichy to me that btcstaking module has so many parameters. At some point it would be good to revisit them a bit.

Yup, this is also my thinking. Up until now we only have added parameters whenever they were needed, now it is time to step back and look at them from system perspective.

PR #582, opened 2024-03-18
fixme: unable to test export of vp dst cache

RafilxTenfen:


x/btcstaking/keeper/genesis_test.go
```diff
@@ -172,5 +172,25 @@ func TestExportGenesis(t *testing.T) {
 		require.Equal(t, evt, evtIdx)
 	}
 
-	// TODO: vp dst cache
+	// since events are needed to test and I cant update the power dist, the check for
+	// power distribution will be done in another gen export
+	// TODO: vp dst cache, comes with empty dst cache :/
```
SebastianElvis, line 177: The reason is that we didn't add covenant signatures to these BTC delegations. `AddDelegation` adds the BTC delegation to the KV store, but it remains as `pending` state. Only after it receives covenant signatures, it will become `active`, thereby having voting power. Covenant signatures are added via `AddCovenantSigs` function

PR #590, opened 2024-03-22
fix: incentive: Use the LCD query path utilised by other modules

vitsalis:


SebastianElvis (approved):
Thanks! CI failed though

PR #593, opened 2024-03-25
Versioned staking params

KonradStaniec:
Fixes: https://github.com/babylonchain/pm/issues/13

Note:
Alternative would be requiring other tools to use archive babylon nodes to check params at some height `n` and then validate/build proper transactions based on params at this height `n`

This approach seems more desirable as params:
- do not change often
- occupy low amount of space in db

This lowers requirements for other tools (covenant, slasher) to full node.

NOTE: 

This approach also makes possible for future extension to recover locked funds. Consider case:
- user locked funds
- we added new covenant key to possible keys via governance.  This operation make user stake incompatible with current params. 
- given that old members are still there, and old parameters were secure we could in theory accept delegation with this staking transaction and validate it against old parameters.

Note-1:
- For now we store all versions in full node, if this proves to be too much, we can always design system to keep on last `n` versions.


x/btcstaking/keeper/params_test.go
```diff
@@ -17,3 +20,72 @@ func TestGetParams(t *testing.T) {
...
+			// Check that params from aggregate query are ok
+			require.EqualValues(t, *generatedParams[i], *allParams[i])
+
+			// Check retrieval by version is ok
+			paramByVersion := k.GetParamsByVersion(ctx, uint32(i))
+			require.NotNil(t, paramByVersion)
+			require.EqualValues(t, *generatedParams[i], *paramByVersion)
+		}
+
+		lastParams := k.GetParams(ctx)
```
gitferry, line 88: we can add the check of `k.GetParamsWithVersion` here

gitferry (approved):
Lgtm!

x/btcstaking/keeper/params.go
```diff
@@ -2,34 +2,134 @@ package keeper
...
+	binary.BigEndian.PutUint32(buf[:], v)
+	return buf[:]
+}
+func uint32FromBytes(b []byte) (uint32, error) {
+	if len(b) != 4 {
+		return 0, fmt.Errorf("invalid uint32 bytes length: %d", len(b))
+	}
+
+	return binary.BigEndian.Uint32(b), nil
+}
```
SebastianElvis, line 26: Could we simply use `byte` to denote a version in order to avoid the conversion thingy?

x/btcstaking/keeper/params.go
```diff
@@ -2,34 +2,134 @@ package keeper
...
+
+func (k Keeper) nextParamsVersion(ctx context.Context) uint32 {
+	paramsStore := k.paramsStore(ctx)
+	it := paramsStore.ReverseIterator(nil, nil)
+	defer it.Close()
+
+	if !it.Valid() {
+		return 0
+	}
```
SebastianElvis, line 49: Probably we might panick here as no parameter found in the system means something really serious happens

x/btcstaking/keeper/params.go
```diff
@@ -2,34 +2,134 @@ package keeper
...
+}
+
+func (k Keeper) getLastParams(ctx context.Context) *types.StoredParams {
+	paramsStore := k.paramsStore(ctx)
+	it := paramsStore.ReverseIterator(nil, nil)
+	defer it.Close()
+
+	if !it.Valid() {
+		return nil
+	}
```
SebastianElvis, line 61: Same here, panic might be better

x/btcstaking/keeper/params.go
```diff
@@ -2,34 +2,134 @@ package keeper
...
+
+func (k Keeper) GetParamsWithVersion(ctx context.Context) (sp types.StoredParams) {
+	lastParams := k.getLastParams(ctx)
+
+	if lastParams == nil {
+		return sp
+	}
+
+	return *k.getLastParams(ctx)
+}
```
SebastianElvis, line 131:
```suggestion
func (k Keeper) GetParamsWithVersion(ctx context.Context) *types.StoredParams {
	return k.getLastParams(ctx)
}
```
Also how about returning a `Params` and a version rather than returning `StoredParams` directly? this way `StoredParams` is entirely internal for `params.go`

x/btcstaking/keeper/msg_server.go
```diff
@@ -321,8 +321,9 @@ func (ms msgServer) CreateBTCDelegation(goCtx context.Context, req *types.MsgCre
 		SlashingTx:       req.SlashingTx,
 		DelegatorSig:     req.DelegatorSlashingSig,
 		UnbondingTime:    uint32(validatedUnbondingTime),
-		CovenantSigs:     nil, // NOTE: covenant signature will be submitted in a separate msg by covenant
-		BtcUndelegation:  nil, // this will be constructed in below code
+		CovenantSigs:     nil,        // NOTE: covenant signature will be submitted in a separate msg by covenant
+		BtcUndelegation:  nil,        // this will be constructed in below code
+		ParamsVersion:    vp.Version, // version of the params against delegations was validated
```
SebastianElvis, line 326: The handler functions `AddCovenantSigs`, `BTCUndelegate`, and `SelectiveSlashingEvidence` need to make use of this version of the parameter for everything, rather than blindly retrieving the latest parameter.

proto/babylon/btcstaking/v1/query.proto
```diff
@@ -19,6 +19,10 @@ service Query {
   rpc Params(QueryParamsRequest) returns (QueryParamsResponse) {
     option (google.api.http).get = "/babylon/btcstaking/v1/params";
   }
+  // ParamsByVersion queries the parameters of the module for a specific version of past params.
+  rpc ParamsByVersion(QueryParamsByVersionRequest) returns (QueryParamsByVersionResponse) {
+    option (google.api.http).get = "/babylon/btcstaking/v1/params/{version}";
+  }
```
SebastianElvis, line 25: Instead of having a new API, how about we simply returning a slice of params in `Params` API?

x/btcstaking/keeper/params.go
KonradStaniec: tbh until we have some decision about how many versions of params we want to store `byte` seems overly limiting i.e only 256 versions.

After this pr  I will create follow up task which may mentions pruning of old versions, after that we will see what type woult fit here.

x/btcstaking/keeper/params.go
KonradStaniec: agreed, just wanted to be consisted what we use in other modules i.e in other modules if params are not found we always return zero value of the struct (which seems weird)

KonradStaniec: althoth this place in particualr has too little context to panic as we may as well setting params in genesis where there is not params exsting yet

x/btcstaking/keeper/params.go
KonradStaniec: tbh I dislike returning tuples from functions as returned values do not have the names, but maybe in this constrained context it makes sense 🤔 will see

proto/babylon/btcstaking/v1/query.proto
KonradStaniec: I avoid aggregate queries if I can in general as then you need to care about pagination and resource usage. In my mind there will be two types of consumers to this api:
- staker which will want latest version to send and validate delegation
- and tools which will want specific versions to do their jobs

I do not see any one interested in getting whole slice of versions. Do I miss something here?

x/btcstaking/keeper/params.go
KonradStaniec: I will add panics in get function. (we should probably have task to improve all modules in this regard)

x/btcstaking/keeper/msg_server.go
KonradStaniec: tbh I wanted to do it in next pr and leave this pr without changing semantic of other messages processing i.e as it currently stand every message is processed as it was before this pr. (against latest version)

But maybe you are right, lets do it here to not forget about it. Then the only follow up task will be potential pruning of old version of params.

KonradStaniec: yup pruning would be tricky, I think the only way to do it correctly would be if we would also added pruning of expired/unbonded delegations, then whole process would be roughly:
1. prune expire/unbonded delegations
2. check if oldest version is used by some delegation, if not then delete it. 

Agree this is not high priority though given size of params and frequency of updates 👍

KonradStaniec: added using correct version in https://github.com/babylonchain/babylon/pull/593/commits/c0bbcf92089a32d6d594f6be2072a5d84bfaad6a

x/btcstaking/keeper/params.go
KonradStaniec: checked how it looks in code and I do not like too much, it makes this two values (params and version) a bit unconnected to each other.

Another way would be to introduce another struct `VersionedParams` which would be used only in operations (not db) but I am not sure it is worth it

cmd/babylond/cmd/genesis.go
```diff
@@ -154,7 +154,8 @@ func PrepareGenesis(
 
 	// btcstaking module genesis
 	btcstakingGenState := btcstakingtypes.DefaultGenesis()
-	btcstakingGenState.Params = genesisParams.BtcstakingParams
+	// here we can start only from single params, which will be initially labelled version 0
+	btcstakingGenState.Params = []*btcstakingtypes.Params{&genesisParams.BtcstakingParams}
```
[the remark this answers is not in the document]
KonradStaniec: Current design assumes for `export/import` that export will return parameters in version order so then index in the collection becomes  version number i.e `exportedParams[0] = Params_with_version_0` 

here we can start only with the one params in the collections so this implicitly becomes version `0`.

If this will prove problematic we can adjust it in future pr's, but this way less code knows about this explicit parameters versioning.

x/btcstaking/keeper/msg_server.go
```diff
@@ -650,13 +661,20 @@ func (ms msgServer) SelectiveSlashingEvidence(goCtx context.Context, req *types.
 	defer telemetry.ModuleMeasureSince(types.ModuleName, time.Now(), types.MetricsKeySelectiveSlashingEvidence)
 
 	ctx := sdk.UnwrapSDKContext(goCtx)
-	bsParams := ms.GetParams(ctx)
 
 	// ensure BTC delegation exists
 	btcDel, err := ms.GetBTCDelegation(ctx, req.StakingTxHash)
```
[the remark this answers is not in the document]
KonradStaniec: hmm I am not sure what you mean.

`btcDel` already containg field:
```
	// version of the params used to validate the delegation
	ParamsVersion uint32 `protobuf:"varint,15,opt,name=params_version,json=paramsVersion,proto3" json:"params_version,omitempty"`
```

so every caller knows what version was used to validate this delegation

Issue #591, opened 2024-03-25
BIP-32 based EOTS randomness derivation/timestamping


SebastianElvis:
Proposal: https://babylon-chain.atlassian.net/wiki/spaces/BABYLON/pages/214630413/EOTS+randomness+derivation+protocol+from+BIP-32

Steps for implementation:

- Babylon side
  - [x] master secret/public randomness, DB schema, and message formats (#594 )
  - [x] Removing `MsgCommitPubRandList` and using master public randomness for verifying finality signatures (#595 )
  - [x] BTC timestamping master public randomness for finality providers (#598 )
  - [x] Docs and changelogs (#604 )
- Finality provider side
  - [x] master secret/public randomness, DB schema, and message formats in EOTS manager
  - [x] master secret/public randomness, DB schema, and message formats in finality provider
  - [x] Removing `MsgCommitPubRandList` and using master public randomness for verifying finality signatures
  - [x] BTC timestamping master public randomness for finality providers
  - [x] Docs and changelogs
- [ ] Bump vigilantes

PR #594, opened 2024-03-26
finality: BIP-32 based randomness derivation

SebastianElvis:
First step towards implementing #591 

This PR introduces BIP-32 based randomness derivation. This includes

- Implementing master secret/public randomness and relevant functions
- Adding master public randomness to `FinalityProvider` DB object
- Adding master public randomness to `MsgCreateFinalityProvider` and modifying the corresponding message handler
- Adding master public randomness to API types

The PR also includes/modifies relevant fuzz/e2e tests.

The subsequent steps include the following and will be done in subsequent PRs

- replacing public randomness committed from `MsgCommitPubRandList` with derived ones
- BTC timestamping master public randomness
-  Allowing finality providers to update master public randomness (TBD)
- Docs and changelogs


proto/babylon/btcstaking/v1/btcstaking.proto
```diff
@@ -25,14 +25,16 @@ message FinalityProvider {
     bytes btc_pk = 4 [ (gogoproto.customtype) = "github.com/babylonchain/babylon/types.BIP340PubKey" ];
     // pop is the proof of possession of babylon_pk and btc_pk
     ProofOfPossession pop = 5;
+    // master_pub_rand is the master public randomness of the finality provider
+    string master_pub_rand = 6;
```
gitferry, line 29: here we use `string` type but in `FinalityProviderWithMeta` we use `bytes` type. Maybe we can make it consistent with `string`?

x/btcstaking/client/cli/tx.go
```diff
@@ -108,13 +108,17 @@ func NewCreateFinalityProvicerCmd() *cobra.Command {
...
-				BabylonPk:   &babylonPK,
-				BtcPk:       btcPK,
-				Pop:         pop,
+				Signer:        clientCtx.FromAddress.String(),
+				Description:   &description,
+				Commission:    &rate,
+				BabylonPk:     &babylonPK,
+				BtcPk:         btcPK,
+				Pop:           pop,
+				MasterPubRand: mpr,
```
gitferry, line 121: should we validate the format of the pub rand?

gitferry (approved):
Great work!

proto/babylon/btcstaking/v1/btcstaking.proto
SebastianElvis: nice catch! made it as `string` now

crypto/eots/randomness.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+func NewMasterRandPair(randSource io.Reader) (*MasterSecretRand, *MasterPublicRand, error) {
+	// get random seed
+	var seed [32]byte
+	if _, err := io.ReadFull(randSource, seed[:]); err != nil {
+		return nil, nil, err
+	}
+	// generate new master key pair
+	masterSK, err := hdkeychain.NewMaster(seed[:], &chaincfg.MainNetParams)
```
KonradStaniec, line 40: Note on `NewMaster` says that we should check for specific err:
```
// NOTE: There is an extremely small chance (< 1 in 2^127) the provided seed
// will derive to an unusable secret key.  The ErrUnusable error will be
// returned if this should occur, so the caller must check for it and generate a
// new seed accordingly.
```

also why hardcoding ` &chaincfg.MainNetParams` ?

proto/babylon/btcstaking/v1/query.proto
```diff
@@ -318,16 +318,18 @@ message FinalityProviderResponse {
   bytes btc_pk = 4 [ (gogoproto.customtype) = "github.com/babylonchain/babylon/types.BIP340PubKey" ];
   // pop is the proof of possession of babylon_pk and btc_pk
   ProofOfPossession pop = 5;
+  // master_pub_rand is the master public randomness of the finality provider
+  string master_pub_rand = 6;
```
KonradStaniec, line 322: I would add comments about the fact that string is base58 encoding

KonradStaniec (commented):
just some late comments, nothing critica :) great work 👍

crypto/eots/randomness.go
SebastianElvis: Good point, will fix in a follow up

> also why hardcoding  &chaincfg.MainNetParams ?

No particular reason tbh. The value is not critical for our use case, so I picked up mainnet. Probably we can leave it as is?

proto/babylon/btcstaking/v1/query.proto
SebastianElvis: Good point, will add in a follow up

crypto/eots/randomness.go
SebastianElvis: resolved in [`b4b486f` (#595)](https://github.com/babylonchain/babylon/pull/595/commits/b4b486f1e98a124341c461984d3065d52dc83df7)

proto/babylon/btcstaking/v1/query.proto
SebastianElvis: resolved in [`b4b486f` (#595)](https://github.com/babylonchain/babylon/pull/595/commits/b4b486f1e98a124341c461984d3065d52dc83df7)

PR #597, opened 2024-03-27
fix: fix non-determinism in voting power rotation

SebastianElvis:
There is a non-determinism bug in voting power rotation, where we iterate over the `activeBTCDels` map, which is not deterministic. This PR fixes the bug by first sorting the keys before iterating over them.

- Reproduce: https://github.com/babylonchain/babylon/tree/3d4b194e020aea069c372b33528a4d12f6e88c1f
- Fix and fuzz test: https://github.com/babylonchain/babylon/tree/ed65719b9d1825a3eb1992b56e2c8d8eecc74957

### DIFF
--- a/x/btcstaking/keeper/power_dist_change.go
+++ b/x/btcstaking/keeper/power_dist_change.go
@@ -2,6 +2,7 @@ package keeper
 
 import (
 	"context"
+	"sort"
 
 	"cosmossdk.io/store/prefix"
 	bbn "github.com/babylonchain/babylon/types"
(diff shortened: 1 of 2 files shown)


x/btcstaking/keeper/power_dist_change.go
```diff
@@ -192,7 +193,16 @@ func (k Keeper) processAllPowerDistUpdateEvents(
 	/*
 		process new BTC delegations under new finality providers in activeBTCDels
 	*/
-	for fpBTCPKHex, fpActiveBTCDels := range activeBTCDels {
+	// sort new finality providers in activeBTCDels to ensure determinism
+	fpBTCPKHexList := []string{}
```
gitferry, line 197:
```suggestion
	fpBTCPKHexList := make([]string, 0, len(activeBTCDels))
```

gitferry (approved):
Nice fix!

PR #598, opened 2024-03-27
finality: BTC timestamping master public randomness

SebastianElvis:
3rd step for #591 

This PR introduces BTC timestamping to finality providers. It enforces one more verification rule over finality signatures: the epoch when the finality provider registers has to be finalised by BTC timestamping. This ensures that the finality provider commits a unique master public randomness, thus preventing committing conflicted master public randomness in two forks.

The PR includes the following changes

- adding a new field `registered_epoch` to the finality provider object. It is the epoch when the finality provider is registered.
- moving the functionality of remembering last finalised epoch from Zone Concierge to Checkpointing module. This is because now both Zone Concierge and BTC staking modules are using the last finalised epoch.
- enforcing the BTC timestamp for finality providers in `AddFinalitySig` msg handler.
- fixing all relevant fuzz/e2e tests.


x/checkpointing/keeper/keeper.go
```diff
@@ -395,3 +399,29 @@ func (k Keeper) GetTotalVotingPower(ctx context.Context, epochNumber uint64) int
 func (k Keeper) GetPubKeyByConsAddr(ctx context.Context, consAddr sdk.ConsAddress) (cmtprotocrypto.PublicKey, error) {
 	return k.epochingKeeper.GetPubKeyByConsAddr(ctx, consAddr)
 }
+
+// GetFinalizedEpoch gets the last finalised epoch
+func (k Keeper) GetFinalizedEpoch(ctx context.Context) (uint64, error) {
```
KonradStaniec, line 404: nitpick: maybe ` GetFinalizedEpoch`  -> `LastFinalizedEpoch` ? 

My mind when sees ` GetFinalizedEpoch`, always internally asks `but which finalized epoch ? ` and I need need to see docstring that we are talking about last one.

KonradStaniec (approved):
Just a minor nit. Great work! Good idea with moving tracking of last finalized epoch to checkpointing module.

x/btcstaking/types/errors.go
```diff
@@ -10,7 +10,7 @@ var (
 	ErrBTCDelegatorNotFound         = errorsmod.Register(ModuleName, 1101, "the BTC delegator is not found")
 	ErrBTCDelegationNotFound        = errorsmod.Register(ModuleName, 1102, "the BTC delegation is not found")
 	ErrFpRegistered                 = errorsmod.Register(ModuleName, 1103, "the finality provider has already been registered")
-	ErrFpAlreadySlashed             = errorsmod.Register(ModuleName, 1104, "the finality provider has already been slashed")
+	ErrFpNotUsable                  = errorsmod.Register(ModuleName, 1104, "the finality provider is not usable (i.e., hasn't been BTC timestamped yet or is slashed)")
```
gitferry, line 13: why remove `ErrFpAlreadySlashed`? We can just add one more error like `ErrFpNotTimestamped` right?

x/btcstaking/keeper/keeper.go
```diff
@@ -67,3 +70,7 @@ func (k Keeper) BeginBlocker(ctx context.Context) error {
 
 	return nil
 }
+
+func (k Keeper) GetFinalizedEpoch(ctx context.Context) (uint64, error) {
```
gitferry, line 74: nit: `GetFinalizedEpoch` -> `GetLastFinalizedEpoch`

x/checkpointing/types/keys.go
```diff
@@ -31,6 +31,8 @@ var (
 
 	AddrToBlsKeyPrefix = append(RegistrationPrefix, 0x0) // where we save the concrete BLS public keys
 	BlsKeyToAddrPrefix = append(RegistrationPrefix, 0x1) // where we save BLS key set
+
+	FinalizedEpochKey = []byte{0x04} // FinalizedEpochKey defines the key to store the last finalised epoch
```
gitferry, line 35:
```suggestion
	LastFinalizedEpochKey = []byte{0x04} // LastFinalizedEpochKey defines the key to store the last finalized epoch
```

gitferry (approved):
LGTM!

PR #427, opened 2024-01-24
E2e test improvements

maurolacy:
Noticed some issues with the BTCStaking tests failing from time to time:
https://app.circleci.com/pipelines/github/babylonchain/babylon/1891/workflows/5e99488b-f864-45ad-856b-a01540cacb0c.

Turns out this is (likely / probably) a wasmer bug, with a [workaround](https://github.com/wasmerio/wasmer/issues/1568#issuecomment-1868697116) that was reported some time ago by Injective.

Changed the babylon docker image to use Debian bookworm instead of Alpine Linux. Seems to be working ~~(though not in CI)~~.

Also, did some syntax fixes, and robustness improvements of BTC Staking checks.

PR #611, opened 2024-04-08
enable static linking

bert2002:
Enable static linking and unify Dockerfile from finality-provider.


KonradStaniec:
Hey, thanks for first contribution.

The thing it though, that we deliberately stopped using `alpine` due to issue with wasm - https://github.com/babylonchain/babylon/pull/427

The issue is now created upstream-  https://github.com/wasmerio/wasmer/issues/4488 and we may return to using alpine after it is fixed upstream.

Issue #613, opened 2024-04-15
Identify inactive finality providers


gitferry:
User story 1 of epic https://github.com/babylonchain/pm/issues/46

* Introduce finality vote tracker to track finality vote histories
* Implement the algorithm to mark inactive finality providers as jailed according to their voting histories
* Add Unjail message to allow users to unjail inactive finality providers

PR #614, opened 2024-04-15
feat: Add finality provider signing info

gitferry:
This PR is the first step towards implementing addressing #613.

This PR added additional parameters and a signing tracker to the `x/finality` module. This is mostly adapted from the [x/slashing](https://docs.cosmos.network/v0.50/build/modules/slashing#states) module of Cosmos-sdk. This PR also added the `jailed` attribute to `FinalityProvider`.

The subsequent PR will implement the liveness handler within `x/finality`'s BeginBlock to update the keeper state and jail inactive finality providers


proto/babylon/finality/v1/params.proto
```diff
@@ -2,10 +2,26 @@ syntax = "proto3";
...
 option go_package = "github.com/babylonchain/babylon/x/finality/types";
 
 // Params defines the parameters for the module.
 message Params {
   option (gogoproto.goproto_stringer) = false;
+  // signed_blocks_window defines the size of the sliding window for tracking finality provider liveness
+  int64 signed_blocks_window  = 1;
+  // min_signed_per_window defines the minimum number of blocks that a finality provider is required to sign
+  // within the sliding window to avoid being jailed
+  bytes min_signed_per_window = 2 [
```
RafilxTenfen, line 18: question: why the `min_signed_per_window` is not an int as well?

x/finality/keeper/keeper.go
```diff
@@ -23,6 +23,11 @@ type (
 		// the address capable of executing a MsgUpdateParams message. Typically, this
 		// should be the x/gov module account.
 		authority string
+
+		// FinalityProviderSigningTracker key: BIP340PubKey bytes | value: FinalityProviderSigningInfo
+		FinalityProviderSigningTracker collections.Map[[]byte, types.FinalityProviderSigningInfo]
+		// FinalityProviderMissedBlockBitmap key: BIP340PubKey bytes | value: byte key for a finality provider's missed block bitmap chunk
+		FinalityProviderMissedBlockBitmap collections.Map[collections.Pair[[]byte, uint64], []byte]
```
SebastianElvis, line 30: I remember Cosmos SDK has memory store dedicated for maintaining KV pairs in memory, as evidenced by `MemStoreKey` variables in our code. Is it possible to use that?

Also I'm not sure whether this should be in memory or DB. If a node restarts for some reason, these data might be lost, and will have different execution results about slashing inactive fps. Will that be a concern?

SebastianElvis (commented):
Great work and solid first step towards jailing mechanism!

Concept ACK, some initial comments

proto/babylon/finality/v1/finality.proto
```diff
@@ -39,3 +41,18 @@ message Evidence {
     // where finality signature is an EOTS signature
     bytes fork_finality_sig = 7 [ (gogoproto.customtype) = "github.com/babylonchain/babylon/types.SchnorrEOTSSig" ];
 }
+
+// FinalityProviderSigningInfo defines a finality provider's signing info for monitoring their
+// liveness activity.
+message FinalityProviderSigningInfo {
```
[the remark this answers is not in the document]
gitferry: We can define it in `btcstaking` but  I think it's more relevant to `finality` as `finality` module is the one that handles finality votes.

proto/babylon/finality/v1/params.proto
gitferry: This is adapted from the cosmos-sdk's [slashing module](https://github.com/cosmos/cosmos-sdk/blob/0cf0c285ea90fe1f321ca63762c1adf86b0c7dc2/x/slashing/proto/cosmos/slashing/v1beta1/slashing.proto#L42). I guess the reason `bytes` is needed is because of `(amino.dont_omitempty) = true`

x/finality/keeper/keeper.go
gitferry: This is adapted from the [x/slasher](https://github.com/cosmos/cosmos-sdk/blob/e46d526b46d6a932b6b409d90894086ce6cd34aa/x/slashing/keeper/keeper.go#L33) module of Cosmos-sdk and it is in DB as it is initiated from `storeService corestoretypes.KVStoreService`

SebastianElvis: I see, thanks for the pointer. This looks like a smart way of using the DB where we we can access the DB like accessing a map in memory. Good to know :+1:

proto/babylon/btcstaking/v1/btcstaking.proto
```diff
@@ -38,6 +38,8 @@ message FinalityProvider {
     // the finality provider is slashed.
     // if it's 0 then the finality provider is not slashed
     uint64 slashed_btc_height = 9;
+    // jailed defines whether the finality provider has been jailed
+    bool jailed = 10;
```
SebastianElvis, line 42: Since we now have this, should we filter out jailed finality provider in `x/btcstaking/keeper/power_dist_change.go`?

x/finality/keeper/liveness.go
```diff
@@ -0,0 +1,11 @@
+package keeper
+
+import (
+	"context"
+)
+
+// HandleLiveness checks liveness of each finality provider from the active set and
+// jail inactive ones
+func (k Keeper) HandleLiveness(ctx context.Context) {
```
SebastianElvis, line 9:
```suggestion
func (k Keeper) CheckLiveness(ctx context.Context) {
```

SebastianElvis (approved):
Solid first step towards the liveness slashing!

proto/babylon/btcstaking/v1/btcstaking.proto
gitferry: yep, for this we need a new `EventPowerDistUpdate` event, correct? Will do in a subsequent PR

x/finality/keeper/liveness.go
gitferry: Prefer `Handle` because what it does is beyond checking. It also needs to take actions (jailing) upon inactivities

PR #616, opened 2024-04-15
chore: constructor function for master secret/public randomness

SebastianElvis:
Needed by https://github.com/babylonchain/finality-provider/pull/284

### DIFF
--- a/crypto/eots/randomness.go
+++ b/crypto/eots/randomness.go
@@ -31,6 +31,11 @@ func RandGen(randSource io.Reader) (*PrivateRand, *PublicRand, error) {
 	return &pk.Key, &j.X, nil
 }
 
+// NewMasterRandPairFromSeed deterministically generates a pair of
+// master/secret randomness from a given seed.
+// NOTE: BIP-32 allows to use seed with 16-64 bytes. We used 32 here
+// as the seed is typically a 32-byte hash, e.g., finality provider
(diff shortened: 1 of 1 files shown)


crypto/eots/randomness.go
```diff
@@ -67,6 +62,15 @@ func NewMasterRandPair(randSource io.Reader) (*MasterSecretRand, *MasterPublicRa
 	return &MasterSecretRand{masterSK}, &MasterPublicRand{masterPK}, nil
 }
 
+func NewMasterRandPair(randSource io.Reader) (*MasterSecretRand, *MasterPublicRand, error) {
+	// get random seed
+	var seed [32]byte
```
gitferry, line 67: Does the seed have to be exactly 32 bytes? Some docstring would be helpful

gitferry (approved):

crypto/eots/randomness.go
SebastianElvis: Good point. Actually it does not need to be 32 bytes, but rather any value between 16-64 bytes ([ref](https://github.com/btcsuite/btcd/blob/master/btcutil/hdkeychain/extendedkey.go#L653-L656)). Used 32 as peripheral programs typically use a hash as the seed, e.g., the finality provider generates 32-byte seed for deriving master randomness pair.

PR #617, opened 2024-04-16
chore: create finality provider to use in phase-1

RafilxTenfen:


RafilxTenfen:
done at finality provider repo https://github.com/babylonchain/finality-provider/pull/288

PR #625, opened 2024-04-22
chore: unifying usage of verification functions for btc staking txs

SebastianElvis:
This PR unifies the usage of verification functions for various txs in BTC staking, and cleans up some of the APIs.


KonradStaniec (approved):
yup it makes sense to remove this function.

PR #633, opened 2024-05-07
hotfix: Remove broadcasting btc timestamps

KonradStaniec:
- When we have a phase-2 IBC channel connection, we are forwarding Bitcoin timestamps and proofs for them
- To generate the proof, we are switching the context to a past block height in order to retrieve the data for a prior state
- Validators that have a high pruning policy (i.e. only keep very few recent blocks), can’t switch to the past state to generate the data, leading to non-determinism

The hotfix involves removing broadcasting of BTC timestamps. We are going to implement a longer term solution in which we do not get historical states for the generation of timestamps for the next testnet. While this is not consensus breaking, this happened only on a recent block, so if we roll back to that block and apply this it consensus won't break for new participants on the ledger.


vitsalis (approved):
LGTM, let's wait for @filippos47 to verify the hotfix before we make a new release.

PR #634, opened 2024-05-07
Hot fix broadcast - merge back to dev

KonradStaniec:


SebastianElvis (dismissed):

SebastianElvis:
The commit for commenting out timestamps seems different compared to the one in `main`? https://github.com/babylonchain/babylon/tree/main

KonradStaniec:
hmm on main it is this commit - https://github.com/babylonchain/babylon/tree/d4a70826d919541ed87fa3539fe0debfb900ddee, the commit after that is merge commit, which will be different on dev

SebastianElvis (approved):

PR #636, opened 2024-05-08
zoneconcierge: avoid querying historical state by generating proofs upon sealing epochs

### LINKED ISSUE #635: zoneconcierge: save BTC timestamp proofs upon sealing an epoch
This is to avoid querying states at a prior height, which might not be compatible with pruning.

- hook upon epoch is sealed
- save proof that CZ header is in an epoch in each epoch's chain info
- new KV store where key is sealed epoch and value is the proofs
- generate and save proof that an epoch is sealed upon sealing this epoch

SebastianElvis:
This PR fixes #635 by generating and saving proofs upon an epoch is sealed, such that Babylon no longer requires nodes to not prune too many recent blocks. 

This includes

- adding hook upon epoch is sealed
- saving proof that CZ header is in an epoch in each epoch's chain info
- new KV store where key is sealed epoch and value is the proof that the epoch is sealed
- generate and save proof that an epoch is sealed upon sealing this epoch


KonradStaniec (commented):
Code wise it looks good !

One thing though, I wanted to verify if everything works as expected mostly for case when somebody prunes everything. 

So I have tweaked nodes in e2e test to have pruning option `everything` and test seemed to pass.  I was mostly interested in `TestBTCTimestampingPhase2RlyTestSuite` testcase.

On the other hand when I looked at logs in e2e, I saw this errors messages:
```
11:54AM INF packet acknowledged dst_channel=channel-0 dst_port=wasm.bbn14hj2tavq8fpesdwxxcu44rty3hh90vhujrvcmstl4zr3txmfvw9sw76fy2 module=x/ibc/channel sequence=2 src_channel=channel-0 src_port=zoneconcierge
11:54AM ERR received an Acknowledgement error message error="invalid packet: Generic error: failed to initialize Babylon epoch: The epoch is not sealed by the epoch's validator set: checkpoint's block_hash (83ce19d759113cdebd2cf26eb41d5e2b9addb9b5c629413f516ce88d0003d525) is not equal to epoch's sealer_block_hash ()" module=x/ibc-zoneconcierge
```
Do you think this is some problem or rather our e2e tests being not perfect ?

SebastianElvis:
> One thing though, I wanted to verify if everything works as expected mostly for case when somebody prunes everything.
> 
> So I have tweaked nodes in e2e test to have pruning option `everything` and test seemed to pass. I was mostly interested in `TestBTCTimestampingPhase2RlyTestSuite` testcase.
> 
> On the other hand when I looked at logs in e2e, I saw this errors messages:
> 
> ```
> 11:54AM INF packet acknowledged dst_channel=channel-0 dst_port=wasm.bbn14hj2tavq8fpesdwxxcu44rty3hh90vhujrvcmstl4zr3txmfvw9sw76fy2 module=x/ibc/channel sequence=2 src_channel=channel-0 src_port=zoneconcierge
> 11:54AM ERR received an Acknowledgement error message error="invalid packet: Generic error: failed to initialize Babylon epoch: The epoch is not sealed by the epoch's validator set: checkpoint's block_hash (83ce19d759113cdebd2cf26eb41d5e2b9addb9b5c629413f516ce88d0003d525) is not equal to epoch's sealer_block_hash ()" module=x/ibc-zoneconcierge
> ```
> 
> Do you think this is some problem or rather our e2e tests being not perfect ?

Nice catch! I have bumped Babylon contract and now the acknowledgment becomes successful. It is because the babylon contract is not the updated version. Please take a look again

KonradStaniec (approved):
>We could add more assertions in timestamping integration e2e in subsequent PRs.
Agreed this can be done in separate pr's

PR #646, opened 2024-05-15
bump Cosmos SDK and wasmd

SebastianElvis:


gitferry:
Now the CI has passed. Did some fixes:
* TestPrepareProposalAtVoteExtensionHeight fixed in [673110d](https://github.com/babylonchain/babylon/pull/646/commits/673110d6efcba5170fc127bd549a1054b1b2db6c)
  * The header info needs `chainID`
  * We need to ensure that BLS signatures are all signed over the most weighted block hash
* FuzzDeterminismBtcstakingBeginBlocker fixed in [812a57d](https://github.com/babylonchain/babylon/pull/646/commits/812a57d389f67898b5bcb13ad5bfab868de909ef)
  * In a [fix](https://github.com/cosmos/cosmos-sdk/pull/19338/files) of Cosmos `v0.50.4`, the `headerInfo` is set to the finalized state via block header
  * In the helper function, the block header is generated randomly. That's why the test won't pass.

go.mod
```diff
@@ -3,67 +3,64 @@ go 1.21
...
+	github.com/CosmWasm/wasmd v0.51.0
 	github.com/btcsuite/btcd v0.24.0
-	github.com/cometbft/cometbft v0.38.5
+	github.com/cometbft/cometbft v0.38.6
 	github.com/cometbft/cometbft-db v0.9.1 // indirect
-	// TODO: the current cosmos-sdk version is a non-released version, which is
-	// to fix the gasLimit problem
-	// should be replaced by a released version if there's any
-	github.com/cosmos/cosmos-sdk v0.50.4-0.20240126152601-c4a2fe2b8987
+	github.com/cosmos/cosmos-sdk v0.50.6
```
maurolacy, line 10: 🎉

wasmbinding/test/custom_query_test.go
```diff
@@ -10,7 +10,7 @@ import (
 
 	"cosmossdk.io/math"
 	"github.com/CosmWasm/wasmd/x/wasm/keeper"
-	wasmvmtypes "github.com/CosmWasm/wasmvm/types"
+	wasmvmtypes "github.com/CosmWasm/wasmvm/v2/types"
```
maurolacy, line 13: As mentioned elsewhere, is this `v2/` really needed?

SebastianElvis: Yes, as wasmd v0.51.0 depends on wasmvm v2, and therefore `go.mod` of Babylon will only have v2. importing `"github.com/CosmWasm/wasmvm/types"` will not compile

x/checkpointing/vote_ext_test.go
```diff
@@ -101,8 +101,9 @@ func FuzzAddBLSSigVoteExtension_InvalidBLSSig(f *testing.F) {
 	})
 }
 
-// FuzzAddBLSSigVoteExtension_EmptyVoteExtensions tests resilience against
-// empty vote extensions
+// FuzzAddBLSSigVoteExtension_EmptyVoteExtensions tests proposals with empty
+// or invalid vote extensions will be rejected PrepareProposal and ProcessProposal.
+// NOTE: empty vote extension will be rejected at VerifyVoteExtension step now
```
gitferry, line 106: Is it true? IIRC this is a comet bft [issue](https://github.com/cometbft/cometbft/issues/2361) where an invalid vote extension could still sneak into the proposal

x/checkpointing/vote_ext_test.go
```diff
@@ -113,27 +114,22 @@ func FuzzAddBLSSigVoteExtension_EmptyVoteExtensions(f *testing.F) {
...
 		for i := uint64(0); i < interval-2; i++ {
-			_, err := helper.ApplyEmptyBlockWithSomeEmptyVoteExtensions(r)
+			_, err := helper.ApplyEmptyBlockWithVoteExtension(r)
 			require.NoError(t, err)
 		}
 		// height 11, i.e., 1st block of next epoch
+		// proposal will be rejected as it contains empty vote extension
+		// since Cosmos SDK v0.50.5 proposal with empty/invalid vote extension
+		// will be rejected by both PrepareProposal and ProcessProposal
 		_, err = helper.ApplyEmptyBlockWithSomeEmptyVoteExtensions(r)
```
gitferry, line 131: Don't think Cosmos SDK 0.50.5 will reject proposal with empty/invalid vote extension because ABCI will not verify the vote extensions because verification is specific to the application.

I debugged into this and found that the reason the proposal is rejected is because the vote extension is not sorted by voting power

gitferry (commented):
Awesome work! Thanks for taking the initative. Left one major comment about vote extension verification in `PrepareProposal`.

app/modules.go
```diff
@@ -1,27 +1,45 @@
...
 func (app *BabylonApp) GetIBCKeeper() *ibckeeper.Keeper {
-	return app.IBCKeeper // This is a *ibckeeper.Keeper
+	return app.IBCKeeper
 }
 
 func (app *BabylonApp) GetScopedIBCKeeper() capabilitykeeper.ScopedKeeper {
 	return app.ScopedIBCKeeper
 }
 
+func (app *BabylonApp) GetBankKeeper() bankkeeper.Keeper {
```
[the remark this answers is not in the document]
SebastianElvis: These are needed for satisfying the `ibctesting.ChainApp` [interface](https://github.com/babylonchain/babylon/blob/1a3316f06a46bcd42ca4ad933641000fc489a3d1/app/modules.go#L14-L15). This allows us to test BTC staking integration by using the ibctesting library

x/checkpointing/vote_ext_test.go
SebastianElvis: I tried to debug into this at https://github.com/babylonchain/babylon/tree/debug-empty-ve, where I reverted to the previous test and sorted the vote extensions again in `ApplyEmptyBlockWithSomeEmptyVoteExtensions`. The test still fails, and the error happens [here](https://github.com/babylonchain/babylon/blob/1188ffdf5f91546de1e78798315dbabf06ae9f78/x/checkpointing/proposal.go#L272-L273) since `PrepareProposal` [fails](https://github.com/babylonchain/babylon/blob/1188ffdf5f91546de1e78798315dbabf06ae9f78/x/checkpointing/proposal.go#L83-L84) with the error `errors.errorString {s: "extended commit vote address  is duplicated"}`. After fixing it in this [commit](https://github.com/babylonchain/babylon/commit/e0d5d9d5b10fec03c139f3b58abba7579face0d7) the error becomes `"failed to verify validator 5047FE3DF427F7E21C18E23621428D055EDA6C43 vote extension signature"`. 

So it seems that in the current version, an honest proposer will reject to propose any proposal with invalid vote extension. Malicious proposer will still sneak invalid vote extensions into the proposal but `ProcessProposal` will reject it due to the absense of checkpoiny tx, but not empty vote extension.

So maybe the proper thing is to revise the statement that "the invalid/empty vote extension will be rejected by PrepareProposal" (but not `ProcessProposal`) and then say the ProcessProposal rejects the proposal since checkpoint tx is empty. Wdyt?

gitferry: The cosmos-sdk will not verify the vote extension, but it will verify the vote extension signature to ensure that the vote extension is created by the correct validator.

I think the reason why the verification of the vote extension signature cannot pass is because we changed the vote extension after it is signed https://github.com/babylonchain/babylon/blob/e0d5d9d5b10fec03c139f3b58abba7579face0d7/testutil/helper/gen_blocks.go#L409

SebastianElvis: > I think the reason why the verification of the vote extension signature cannot pass is because we changed the vote extension after it is signed

You are right. Is this expected as the function's name suggests that some of the vote extensions are empty but it actually sets random values (rather than empty values) to vote extensions?

Given that in this function, some of the vote extensions are invalid, this function is doomed to fail to provide a checkpoint at the epoch boundary, as per the verifications imposed by `PrepareProposal`.

gitferry: Yeah calling it `EmptyVoteExtension` is probably not appropriate. Maybe `ArbitraryVoteExtension` is better because the purpose of this test is to test the case where even some of arbitrary vote extensions are sneaked in, the proposal should be successful (we cannot prevent them so we ignore them).

I think a proper fix is to let arbitrary vote extensions inserted before they are signed.

x/checkpointing/proposal.go
```diff
@@ -170,6 +171,14 @@ func (h *ProposalHandler) getValidBlsSigs(ctx sdk.Context, extendedVotes []abci.
 			h.logger.Error("failed to unmarshal vote extension", "err", err)
 			continue
 		}
+
+		if !bytes.Equal(*ve.BlockHash, blockHash) {
```
KonradStaniec, line 175: hm this seems like a important fix it should probably go into separate pr with separate tests, but yep lets not block this pr with that.

x/checkpointing/vote_ext_test.go
KonradStaniec: from what I am aware this is still issue.

KonradStaniec (approved):
lgtm! thanks for doing this update!

x/checkpointing/vote_ext_test.go
SebastianElvis: OK debugged a bit and found that

- ProcessProposal seems no longer resililent against empty vote extensions. The `validateExtendedCommitAgainstLastCommit` function verifies whether the number of vote extensions is same as in the local last commit information. This means the proposer can simply drop some of the vote extensions in the proposal, such that [ValidateVoteExtensions](https://github.com/babylonchain/babylon/blob/2cff49afa3cbb781077d673cdd3ad63a3c7e9592/x/checkpointing/proposal.go#L289-L294) will fail, and the node will panick and recover.
- Our test methodology might have some issue. If we comment out things in [epoch.IsVoteExtensionProposal(h.Ctx)](https://github.com/babylonchain/babylon/blob/2cff49afa3cbb781077d673cdd3ad63a3c7e9592/testutil/helper/gen_blocks.go#L416-L422), then the error message becomes `errors.errorString {s: "extended commit votes length 10 does not match last commit votes length 0"}`. This is because the `ctx` inside `ProcessProposal` does not have the votes locally. It's not straightforward to pass the votes into `ctx` since it's hidden inside enclosures and private functions in base app.

Any ideas?

SebastianElvis:
Thanks @gitferry @KonradStaniec for the comments! You guys are right that the proposal could still contain invalid vote extensions and we have to be resilient against those. I have fixed the tests, and along the way we have found a bunch of new things:

- In `ProcessProposal`, we used to return err if `ValidateVoteExtensions` fails, and returning error will make the node panick. Given that now it validates a few more things that could be manipulated by the adversarial proposer, we should simply reject the proposal without panic.
- `ProcessProposal` now also validates the `ProposedLastCommit` given in the request due to `ValidateVoteExtensions`. This PR has added relevant mock code to construct `ProposedLastCommit`.
- `FuzzAddBLSSigVoteExtension_SomeInvalidVoteExtensions` has to be resilient against invalid vote extensions, but not invalid signatures over vote extensions (upon which the proposal will be rejected). This PR has fixed `ApplyEmptyBlockWithSomeInvalidVoteExtensions` to generate invalid vote extensions but with valid signatures.

KonradStaniec (approved):

PR #648, opened 2024-05-20
Add test vectors for other implementations of scripts

KonradStaniec:
Adding intial implementation of test vectors to test combatiblity between implementations of our scripts. 

Added @SebastianElvis and @gbarkhatov  as reviewers as they are working with rust and js implementation of our scripts.


btcstaking/staking_testvectors_test.go
```diff
@@ -0,0 +1,432 @@
...
+	unbondingTime int,
+	unbondingFee int,
+	magicBytes []byte,
+) string {
+	emptyHash := [32]byte{}
+	eh, err := chainhash.NewHash(emptyHash[:])
+	require.NoError(t, err)
+	covenantKeys := generateKeys(t, numCovenantKeys)
+	finalityKeys := generateKeys(t, numFinalityKeys)
+	stakerKeys := generateKeys(t, 1)
```
SebastianElvis, line 330: Do you think we need to make the test vector generation deterministic, such that everytime we run the test case we get the same vectors?

KonradStaniec: at this point, I do not see a need to do that. This generator at this point is mostly a helper to avoid filling jsons manually.

Downstream consumers do not need to re-generate the testvectores themselves, they need to:
- parse `vectors.json` file
- build staking output/unbonding transaction/scripts based on `parameters` in this json
- check whether locally built objects are the same as `expected`.

btcstaking/staking_testvectors_test.go
```diff
@@ -0,0 +1,432 @@
...
+		panic(fmt.Sprintf("Error decoding JSON: %s", err))
+	}
+
+	return &cases
+}
+
+type Parameters struct {
+	CovenantPublicKeys   []string `json:"covenant_public_keys"`
+	CovenantQuorum       int      `json:"covenant_quorum"`
+	FinalityProviderKeys []string `json:"finality_provider_public_keys"`
```
SebastianElvis, line 80:
```suggestion
	FinalityProviderPublicKeys []string `json:"finality_provider_public_keys"`
```

btcstaking/staking_testvectors_test.go
```diff
@@ -0,0 +1,432 @@
...
+	UnbondingTransactionHex            string `json:"unbonding_transaction_hex"`
+	UnbondingTransactionTimeLockScript string `json:"unbonding_transaction_time_lock_script_hex"`
+	UnbondingTransactionSlashingScript string `json:"unbonding_transaction_slashing_script_hex"`
+	OpReturnScript                     string `json:"op_return_script_hex"`
+}
+
+type TestCase struct {
+	Description string     `json:"name"`
+	Parameters  Parameters `json:"parameters"`
+	Expected    Expected   `json:"expected"`
```
SebastianElvis, line 108:
```suggestion
	Parameters  *Parameters `json:"parameters"`
	Expected    *Expected   `json:"expected"`
```

btcstaking/staking_testvectors_test.go
```diff
@@ -0,0 +1,432 @@
...
+}
+
+type Expected struct {
+	StakingOutputPkScript              string `json:"staking_output_pkscript_hex"`
+	StakingOutputValue                 int    `json:"staking_output_value"`
+	StakingTransactionTimeLockScript   string `json:"staking_transaction_timelock_script_hex"`
+	StakingTransactionUnbondingScript  string `json:"staking_transaction_unbonding_script_hex"`
+	StakingTransactionSlashingScript   string `json:"staking_transaction_slashing_script_hex"`
+	UnbondingTransactionHex            string `json:"unbonding_transaction_hex"`
+	UnbondingTransactionTimeLockScript string `json:"unbonding_transaction_time_lock_script_hex"`
```
SebastianElvis, line 100: Do we need unbonding tx output pk script?

btcstaking/staking_testvectors_test.go
```diff
@@ -0,0 +1,432 @@
...
+
+	// Decode the JSON data into the struct
+	if err := decoder.Decode(&cases); err != nil {
+		panic(fmt.Sprintf("Error decoding JSON: %s", err))
+	}
+
+	return &cases
+}
+
+type Parameters struct {
```
SebastianElvis, line 77: Maybe not in this PR: do we want test vectors for Schnorr/adaptor signatures over staking/slashing/unbonding txs?

SebastianElvis (approved):

btcstaking/staking_testvectors_test.go
KonradStaniec: not really, as we have whole unbonding tx hex which contains inside whole unbonding output along with pkscript

btcstaking/staking_testvectors_test.go
KonradStaniec: I will think about this, this is not that easy as schnorr sigs have this randomness parameter inside which means, we cannot just bytes equality for signatures, we can only check signature/adaptor validity.  I wonder whether it is worth it.

This is definitely something for sparate pr to improve on.

SebastianElvis: > we can only check signature/adaptor validity. I wonder whether it is worth it.

This is needed in BTC staking integration where the consumer side needs to verify the Schnorr/adaptor signatures over these txs received from Babylon. If the signatures are invalid then the consumer is supposed to reject the corresponding BTC delegation.

KonradStaniec: hmm good point. Probably more imporant would be to test adaptor sigs verifcations, as verifing stand schnorr signatures is probably available in libraries in most languages. Created issue to track it- https://github.com/babylonchain/babylon/issues/653

PR #650, opened 2024-05-21
epoching: removing apphash chain in epoching module

SebastianElvis:
Epoching module used to remember the `AppHash` of every block, in order to prove a consumer header is committed to a Babylon header. This is no longer needed, as now we generate proof that a consumer header is committed to the sealer header of an epoch directly. Yet, its overhead is not negligible in terms of computation (upon each epoch we need to generate a Merkle root) and storage (we need to remember every `AppHash`). Thus this PR proposes to remove it.


KonradStaniec (approved):
If it is no longer needed , 👍  for removing it.

PR #655, opened 2024-05-22
finality: Merkle tree based public randomness commitment

SebastianElvis:


x/finality/keeper/msg_server.go
```diff
@@ -97,18 +96,21 @@ func (ms msgServer) AddFinalitySig(goCtx context.Context, req *types.MsgAddFinal
...
-	// verify EOTS signature w.r.t. public randomness
-	fpBTCPK, err := fpPK.ToBTCPK()
-	if err != nil {
+	// verify the proof of inclusion for this public randomness
+	if err := req.VerifyInclusionProof(prCommit.Commitment); err != nil {
 		return nil, err
 	}
-	if err := eots.Verify(fpBTCPK, pubRand.ToFieldVal(), req.MsgToSign(), req.FinalitySig.ToModNScalar()); err != nil {
+	// the public randomness is good, set the public randomness
+	ms.SetPubRand(ctx, req.FpBtcPk, req.BlockHeight, *req.PubRand)
```
gitferry, line 110: We probably don't need to keep the randomness in storage. We can add a TODO here

x/finality/keeper/msg_server.go
```diff
@@ -97,18 +96,21 @@ func (ms msgServer) AddFinalitySig(goCtx context.Context, req *types.MsgAddFinal
...
+
+	// verify EOTS signature w.r.t. public randomness
+	if err := req.VerifyEOTSSig(); err != nil {
```
gitferry, line 113: tbh. I feel a little bit weird to see these verification methods within the requestion. I think it is up to the finality keeper to verify the request content. Otherwise, the verification process is opaque.

Also, I think the two verification processes should be atomic from the verifier's perspective because a seemingly valid EOTS sig makes no sense if the public randomness is not committed before

x/finality/keeper/msg_server.go
```diff
@@ -184,10 +186,11 @@ func (ms msgServer) CommitPubRandList(goCtx context.Context, req *types.MsgCommi
 
 	// ensure the request contains enough number of public randomness
 	minPubRand := ms.GetParams(ctx).MinPubRand
-	givenPubRand := len(req.PubRandList)
+	givenPubRand := req.NumPubRand
 	if uint64(givenPubRand) < minPubRand {
```
gitferry, line 190:
```suggestion
	if givenNumPubRand < minPubRand {
```

x/finality/keeper/msg_server.go
```diff
@@ -198,20 +201,27 @@ func (ms msgServer) CommitPubRandList(goCtx context.Context, req *types.MsgCommi
...
 
+	prCommit := &types.PubRandCommit{
+		StartHeight: req.StartHeight,
+		NumPubRand:  req.NumPubRand,
+		Commitment:  req.Commitment,
+	}
+
+	// get last public randomness commitment
+	// TODO: allow committing public randomness earlier than existing ones?
+	lastPrCommit := ms.GetLastPubRandCommit(ctx, req.FpBtcPk)
```
gitferry, line 212: What would go wrong if we only keep the latest commitment?

gitferry (commented):
Amazing work! Some initial comments and questions:

x/finality/client/cli/tx.go
```diff
@@ -34,8 +35,8 @@ func GetTxCmd() *cobra.Command {
 
 func NewCommitPubRandListCmd() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "commit-pubrand-list [fp_btc_pk] [start_height] [pub_rand1]  [pub_rand2] ... [sig]",
-		Args:  cobra.MinimumNArgs(4),
+		Use:   "commit-pubrand-list [fp_btc_pk] [start_height] [num_pub_rand] [commitment] [sig]",
+		Args:  cobra.MinimumNArgs(5),
```
RafilxTenfen, line 39:
```suggestion
		Args:  cobra.ExactArgs(5),
```

RafilxTenfen (approved):
only one suggestion, but looks good to me

x/finality/keeper/msg_server.go
RafilxTenfen: @gitferry could you elaborate on this part?
> Otherwise, the verification process is opaque.

x/finality/keeper/msg_server.go
KonradStaniec: good point! 

Randomness will become part of stored signature, so no need to store it directly.

x/finality/keeper/msg_server.go
```diff
@@ -97,18 +96,21 @@ func (ms msgServer) AddFinalitySig(goCtx context.Context, req *types.MsgAddFinal
...
 	if err != nil {
-		return nil, types.ErrPubRandNotFound
+		return nil, err
 	}
 
-	// verify EOTS signature w.r.t. public randomness
-	fpBTCPK, err := fpPK.ToBTCPK()
-	if err != nil {
+	// verify the proof of inclusion for this public randomness
+	if err := req.VerifyInclusionProof(prCommit.Commitment); err != nil {
```
KonradStaniec, line 106: shouldn't we also verify that field `req.Proof.Index` and `req.Proof.Total` matches commitment we stored earlier ? (otherwise we could receive valid proofs for different trees ?)

x/finality/keeper/msg_server.go
SebastianElvis: Make sense, made the verification of pub rand commit and finality sig as a single standalone function

x/finality/keeper/msg_server.go
SebastianElvis: We still keep all pub rand commitments, not just the latest one. In the current implementation, if a FP commits to interval `[x, y]`, then it can no longer commit to any interval that is earlier than `x`. I have mixed feeling on this design

- A rational FP will commit incrementally rather than out of order,
- but on the other hand, if a FP mistakenly commits to a height that is very far away, then that means it can never vote for earlier heights.

So whether to allow out of order commitment seems to be a design question. Thoughts? We could do any modification in a follow up PR

x/finality/keeper/msg_server.go
SebastianElvis: Nice catch!

SebastianElvis:
Thanks guys for the comments! I have addressed all of them, most notably verifying proof index and removing public randomness storage. Please take a look again

x/finality/keeper/msg_server.go
KonradStaniec: I think in isolation, it would be harmless to allow commitments for past ranges. 

In my mind it start looking weird when you mix it up with timestamping. Now you can have commitment for early blocks, timestamped in later later epoch. I am not sure whether this introduces any problems, but it will definitely increase implementation complexity (which in my book is itself a problem 😅 )

Therefore I would disallow such out of order commitments, at least for now, and potentially re-visit the problem when we have full flow: timestamping and finality provider side. Maybe during implementation of finality provider side we will hit some issues which will warrants those out-of-order commitments.

x/finality/keeper/msg_server.go
gitferry: @RafilxTenfen for example, `req.VerifyEOTSSig()`, it just verifies itself. You don't know what it verifies exactly unless you look into it.

x/finality/types/msg.go
```diff
@@ -17,48 +16,50 @@ var (
...
+// VerifyFinalitySig verifies the finality signature message w.r.t. the
+// public randomness commitment. The verification includes
+// - verifying the proof of inclusion of the given public randomness
+// - verifying the finality signature w.r.t. the given block height/hash
+func VerifyFinalitySig(m *MsgAddFinalitySig, prCommit *PubRandCommit) error {
+	// verify the index of the public randomness
+	heightOfProof := prCommit.StartHeight + uint64(m.Proof.Index)
+	if m.BlockHeight != heightOfProof {
+		return ErrInvalidFinalitySig.Wrapf("the inclusion proof (for height %d) does not correspond to the given height (%d) in the message", heightOfProof, m.BlockHeight)
+	}
```
KonradStaniec, line 32: shouldn't we also have the check:
```
	if uint64(m.Proof.Total) != prCommit.NumPubRand {
		return ErrInvalidFinalitySig.Wrapf("the total number of public randomnesses in the proof (%d) does not match the number of public randomnesses committed (%d)", m.Proof.Total, prCommit.NumPubRand)
	}

```

KonradStaniec (approved):
one final comment and lgtm! great work!

One final note, I noticed that we store in state only half of the signature:
```
FinalitySig *github_com_babylonchain_babylon_types.SchnorrEOTSSig 
```
and I somehow though we store whole thine ie `(R, s)`, thats why i was in favour of not storing `R` separately. I have a feeling, that it would be beneficial to store whole thing as otherwise other consumers do not have possibility of verifying those halfs. Not for this pr, but something to keep in mind.

x/finality/keeper/msg_server.go
gitferry: Yeah, adding timestamping into consideration complicates the situation. I think a very heigh-level guidline is that for every height that a FP has voting power, it should be able to make a commitment. If we are keeping the commitment order, then we should ensure that a FP will not skip heights that it has voting power when making commiments

x/finality/types/msg.go
```diff
@@ -17,48 +16,50 @@ var (
...
 func (m *MsgAddFinalitySig) MsgToSign() []byte {
 	return msgToSignForVote(m.BlockHeight, m.BlockAppHash)
 }
 
-func (m *MsgAddFinalitySig) VerifyEOTSSig(pubRand *bbn.SchnorrPubRand) error {
+// VerifyFinalitySig verifies the finality signature message w.r.t. the
+// public randomness commitment. The verification includes
+// - verifying the proof of inclusion of the given public randomness
+// - verifying the finality signature w.r.t. the given block height/hash
+func VerifyFinalitySig(m *MsgAddFinalitySig, prCommit *PubRandCommit) error {
```
gitferry, line 27: I'm not sure it's clear enough. Tbh, I would expect this is a keeper method that takes in the ingredients for verification, ie. inclusion proof, pubrand, pk, and sig. This way, the caller will clearly know what exactly the method is verifying. The PubRandCommit is retrieved from the keeper

gitferry (approved):
LGTM! We can fix remaining issues in a separate PR

SebastianElvis:
> that it would be beneficial to store whole thing as otherwise other consumers do not have possibility of verifying those halfs. Not for this pr, but something to keep in mind.

Actually that might be a reason why we should keep the pub rand storage :sweat_smile: 

> one final comment and lgtm! great work!
> 
> One final note, I noticed that we store in state only half of the signature:
> 
> ```
> FinalitySig *github_com_babylonchain_babylon_types.SchnorrEOTSSig 
> ```
> 
> and I somehow though we store whole thine ie `(R, s)`, thats why i was in favour of not storing `R` separately. I have a feeling, that it would be beneficial to store whole thing as otherwise other consumers do not have possibility of verifying those halfs. Not for this pr, but something to keep in mind.

Good point. Actually this is a reason why we should keep the pub rand storage. I added the pub rand storage back for this scenario in case a consumer wants to prove BTC staking finality to others.

x/finality/types/msg.go
SebastianElvis: I prefer to make this function generic as the message is rather self contained. It also allows us to test the verification without using a keeper. We can add a private keeper function that invokes this one but I'd prefer to define fewer levels of abstraction tbh. Can be done in a future PR if we really want to

gitferry: The thing that bugs me is that the current in the current implementation, it's the message that verifies something from a keeper, but the fact should be the other way around. The validity of a signature depends on the state of the keeper, so imo it makes sense to involve the keeper in test. We can mock the getting of PubRandCommit, so I don't think it will make the test complex

PR #652, opened 2024-05-21
feat(finality): Merkle tree-based public randomness commitment

SebastianElvis:

PR #657, opened 2024-05-29
feat: BTC multisig use `sdk.AccAddress`

RafilxTenfen:
- Update to use `sdk.AccAddress` instead of Public Key to identify BTC staker bech32 address
- Proto Changes
  -  Add new `ProofOfPossesionBTC`
  - Update `BTCDelegation` to use staker address instead of bbn pk
  - Update `BTCDelDistInfo` to use staker address instead of bbn pk
  - Update `BTCDelegationResponse` to use staker address
  - Remove signer from `MsgCreateBTCDelegation` and replace bbn pk by staker addr 
- Add `WalletName` to `NodeConfig` in e2e test

Ref https://github.com/babylonchain/pm/pull/38

Other PR's will be created to
- [x] Add e2e test with multisig #664 
- [x] Add e2e test with fee grant #670  
- [x] Update btc-staker https://github.com/babylonchain/btc-staker/pull/167
- [x] Update covenant-emulator https://github.com/babylonchain/covenant-emulator/pull/47


proto/babylon/btcstaking/v1/pop.proto
SebastianElvis, line 1: Is `ProofOfPossession` needed because of the finality provider registration?

gitferry: I guess so. All the finality provider related pop need to be changed. Maybe we can use a base branch from `dev` to accommodate all the changes related to the ADR 20 and do finality provider change in a separate PR. This PR is already a bit long

x/btcstaking/types/pop.go
```diff
@@ -87,6 +110,27 @@ func NewPoPWithECDSABTCSig(babylonSK cryptotypes.PrivKey, btcSK *btcec.PrivateKe
...
+func NewPoPBTCWithECDSABTCSig(addr sdk.AccAddress, btcSK *btcec.PrivateKey) (*ProofOfPossessionBTC, error) {
+	pop := ProofOfPossessionBTC{
+		BtcSigType: BTCSigType_ECDSA,
+	}
+
+	// generate pop.BtcSig = ecdsa_sign(sk_BTC, pop.BabylonSig)
+	// NOTE: ecdsa.Sign has to take the message as string.
+	// So we have to hex babylonSig before signing
+	babylonSigHex := hex.EncodeToString(addr.Bytes())
+	btcSig, err := ecdsa.Sign(btcSK, babylonSigHex)
```
gitferry, line 125:
```suggestion
	// generate pop.BtcSig = ecdsa_sign(sk_BTC, pop.BabylonSig)
	// NOTE: ecdsa.Sign has to take the message as string.
	// So we have to hex addr before signing
	addrHex := hex.EncodeToString(addr.Bytes())
	btcSig, err := ecdsa.Sign(btcSK, addrHex)
```

x/btcstaking/types/pop.go
```diff
@@ -213,13 +325,31 @@ func (pop *ProofOfPossession) VerifyBIP340(babylonPK cryptotypes.PubKey, bip340P
...
+// VerifyBIP340 verifies the validity of PoP where Bitcoin signature is in BIP-340
+// 1. verify(sig=sig_btc, pubkey=pk_btc, msg=staker_addr)?
+func (pop *ProofOfPossessionBTC) VerifyBIP340(stakerAddr sdk.AccAddress, bip340PK *bbn.BIP340PubKey) error {
+	return VerifyBIP340(pop.BtcSigType, pop.BtcSig, bip340PK, stakerAddr.Bytes())
+}
+
+// VerifyBIP340 verifies the validity of PoP where Bitcoin signature is in BIP-340
+// 1. verify(sig=sig_btc, pubkey=pk_btc, msg=pop.BabylonSig)?
+// 2. verify(sig=pop.BabylonSig, pubkey=pk_babylon, msg=pk_btc)?
+func (pop *ProofOfPossession) VerifyBIP340(babylonPK cryptotypes.PubKey, bip340PK *bbn.BIP340PubKey) error {
```
gitferry, line 348: Should we remove `ProofOfPossession` type?

x/btcstaking/types/pop.go
```diff
@@ -361,31 +491,25 @@ func VerifyBIP322SigPop(
 
 // VerifyBIP322 verifies the validity of PoP where Bitcoin signature is in BIP-322
 // after decoding pop.BtcSig to bip322Sig which contains sig and address,
-// 1. verify whether bip322 pop signature where msg=pop.BabylonSig
-// 2. verify(sig=pop.BabylonSig, pubkey=babylonPK, msg=bip340PK)?
-func (pop *ProofOfPossession) VerifyBIP322(babylonPK cryptotypes.PubKey, bip340PK *bbn.BIP340PubKey, net *chaincfg.Params) error {
-	if pop.BtcSigType != BTCSigType_BIP322 {
+// 1. verify whether bip322 pop signature where msg=signedMsg
```
gitferry, line 494:
```suggestion
// verify whether bip322 pop signature where msg=signedMsg
```

x/btcstaking/types/pop.go
```diff
@@ -361,31 +491,25 @@ func VerifyBIP322SigPop(
...
-	// Eventually we need to use tmhash.Sum(pop.BabylonSig) rather than bbnSigHashHexBytes
-	// ref: https://github.com/babylonchain/babylon/issues/433
-	bbnSigHashHexBytes := babylonSigToHexHash(pop.BabylonSig)
-
 	btcKeyBytes, err := bip340PK.Marshal()
 	if err != nil {
 		return err
 	}
 
 	// 1. Verify Bip322 proof of possession signature
```
gitferry, line 510:
```suggestion
	// Verify Bip322 proof of possession signature
```

x/btcstaking/types/pop.go
```diff
@@ -440,3 +608,12 @@ func (pop *ProofOfPossession) ValidateBasic() error {
 
 	return nil
 }
+
+// ValidateBasic checks if there is a BTC Signature.
+func (pop *ProofOfPossessionBTC) ValidateBasic() error {
```
gitferry, line 613: can we further parse the sig to see if it is a correct Schnorr sig in format?

gitferry (commented):
Nice work! Some initial comments:

proto/babylon/btcstaking/v1/pop.proto
SebastianElvis: Yeah things get a bit complicated. Tbh I feel it's also fine to enforce signer of FP registration message to be FP himself. FP registration is also a message sent from users, rather than a protocol-level thing like BTC checkpoints

RafilxTenfen: I agree, the FP should be the signer as well and would have the same PoP as the BTCDelegation 
I will add a task for that

gitferry: Sounds good. I guess ADR also needs update?

x/btcstaking/types/pop.go
RafilxTenfen: will be done in another PR
https://github.com/babylonchain/babylon/issues/663

x/btcstaking/types/pop.go
RafilxTenfen: https://github.com/babylonchain/babylon/pull/657/commits/d9870805efedd9c05d1373da8496830e7021c705

x/btcstaking/types/pop.go
RafilxTenfen: Agree, done at https://github.com/babylonchain/babylon/pull/657/commits/499149171c933203fafee978b048226c3c784ebf

proto/babylon/btcstaking/v1/pop.proto
RafilxTenfen: https://github.com/babylonchain/pm/issues/40

gitferry (commented):
Thanks for the changes. One high-level comment is that I feel this PR does not complete the whole feature described in ADR-020, which seems to require two more PRs. So I would expect a base branch like [this](https://github.com/babylonchain/babylon/pull/652) to collect the whole feature set before merging it to `dev`. This way, this PR is part of the feature and can be renamed to something like `ADR-020: Replace Babylon Pubkey with Staker Address`, [ref](https://github.com/cometbft/cometbft/pull/1154).

SebastianElvis:
> Thanks for the changes. One high-level comment is that I feel this PR does not complete the whole feature described in ADR-020, which seems to require two more PRs. So I would expect a base branch like [this](https://github.com/babylonchain/babylon/pull/652) to collect the whole feature set before merging it to `dev`. This way, this PR is part of the feature and can be renamed to something like `ADR-020: Replace Babylon Pubkey with Staker Address`, [ref](https://github.com/cometbft/cometbft/pull/1154).

Hmm given that this PR is rather self contained (unlike the pub rand commit one where we cannot stay on the status after reverting BIP32 for long), probably this PR is fine to be merged to `dev`?

KonradStaniec:
>One high-level comment is that I feel this PR does not complete the whole feature described in ADR-020, which seems to require two more PRs

From what I see, the only two prs missing are:
- test
- modifications in adjecent repo
If thats all imo we can merge it to dev.

Although lets wait with merging this at least till core meeting. My main concern is, that we would like to return to running devnets, and if we merge it now we would need:
- either run devnet from some past commit, which may be messy as we usually run devnets from top of the dev branch
- or quickly modify other programs for devnet to make sense.

gitferry:
@SebastianElvis @KonradStaniec Makes sense. I thought we also needed to change `pop` in `MsgCreateFinalityProvider`, but I guess that is independent of this PR. Let's merge this PR to `dev` then when we are ready

proto/babylon/btcstaking/v1/pop.proto
```diff
@@ -27,6 +27,17 @@ message ProofOfPossession {
     bytes btc_sig = 3;
 }
 
+// ProofOfPossessionBTC is the proof of possession that a Babylon
+// address and a Bitcoin secp256k1 secret key are held by the same
+// person
+message ProofOfPossessionBTC {
```
SebastianElvis, line 33: We can remove the original `ProofOfPossession` now right?

x/btcstaking/keeper/msg_server.go
```diff
@@ -179,8 +179,13 @@ func (ms msgServer) CreateBTCDelegation(goCtx context.Context, req *types.MsgCre
 	// - is smaller than math.MaxUint16 (due to check in req.ValidateBasic())
 	validatedUnbondingTime := uint16(req.UnbondingTime)
 
+	stakerAddr, err := sdk.AccAddressFromBech32(req.StakerAddr)
+	if err != nil {
+		return nil, types.ErrInvalidStakingTx.Wrapf("invalid staker addr %s: %v", req.StakerAddr, err)
+	}
```
SebastianElvis, line 185: Could this be put under `req.ValidateBasic()`? Also I'm not entirely sure if this check is necessary -- if `StakerAddr` is malformed then Cosmos SDK will reject `req` anyway since the signature verification shall not pass

x/btcstaking/types/pop.go
```diff
@@ -213,13 +325,31 @@ func (pop *ProofOfPossession) VerifyBIP340(babylonPK cryptotypes.PubKey, bip340P
...
+func (pop *ProofOfPossessionBTC) VerifyBIP340(stakerAddr sdk.AccAddress, bip340PK *bbn.BIP340PubKey) error {
+	return VerifyBIP340(pop.BtcSigType, pop.BtcSig, bip340PK, stakerAddr.Bytes())
+}
+
+// VerifyBIP340 verifies the validity of PoP where Bitcoin signature is in BIP-340
+// 1. verify(sig=sig_btc, pubkey=pk_btc, msg=pop.BabylonSig)?
+// 2. verify(sig=pop.BabylonSig, pubkey=pk_babylon, msg=pk_btc)?
+func (pop *ProofOfPossession) VerifyBIP340(babylonPK cryptotypes.PubKey, bip340PK *bbn.BIP340PubKey) error {
+	if err := VerifyBIP340(pop.BtcSigType, pop.BtcSig, bip340PK, pop.BabylonSig); err != nil {
+		return fmt.Errorf("failed to verify possesion of babylon sig by the BTC key: %w", err)
```
SebastianElvis, line 350:
```suggestion
		return fmt.Errorf("failed to verify possession of babylon sig by the BTC key: %w", err)
```

test/e2e/btc_staking_e2e_test.go
```diff
@@ -529,6 +456,163 @@ func (s *BTCStakingTestSuite) Test5SubmitStakerUnbonding() {
...
+		testStakingInfo.SlashingTx,
+		delegatorSig,
+		testUnbondingInfo.UnbondingTx,
+		testUnbondingInfo.SlashingTx,
+		uint16(unbondingTime),
+		btcutil.Amount(testUnbondingInfo.UnbondingInfo.UnbondingOutput.Value),
+		delUnbondingSlashingSig,
+		wGratee,
+		false,
+		fmt.Sprintf("--fee-granter=%s", feePayerAddr.String()),
```
SebastianElvis, line 592: Really good idea on using fee grant :+1: very clean

x/btcstaking/types/pop.go
```diff
@@ -118,12 +162,46 @@ func newPoPWithBIP322Sig[A btcutil.Address](
...
+	pop.BtcSig = bip322SigEncoded
+
+	return &pop, nil
+}
+
+func newPoPBTCWithBIP322Sig[A btcutil.Address](
+	addressToSign sdk.AccAddress,
+	btcSK *btcec.PrivateKey,
+	net *chaincfg.Params,
+	bip322SignFn bip322Sign[A],
```
SebastianElvis, line 178: It seems that the generic type `A` is not used in this function at all. I'm wondering why don't we simply define 

```go
type bip322Sign func(sg []byte,
	privKey *btcec.PrivateKey,
	net *chaincfg.Params) (btcutil.Address, []byte, error)
```

without the generic type?

SebastianElvis (approved):
Great work!

proto/babylon/btcstaking/v1/pop.proto
RafilxTenfen: after we decide on finality provider, if it is removed there too 
we can remove it ^^

x/btcstaking/keeper/msg_server.go
RafilxTenfen: good question, I will test if the SDK already checks that for us

SebastianElvis: It's already called at https://github.com/babylonchain/babylon/pull/657/files/c39fc5f6d825d8ea02407794de5786fc2a62d414#diff-0490899e13cee0223ce9b01147d9ad2753112c4e8118443e1b54684738fc4969R158-R161

x/btcstaking/types/pop.go
RafilxTenfen: Seems like is being used 
https://github.com/babylonchain/babylon/blob/dbdffc6e5328bfa99935507f44a61018e5f4ae7a/x/btcstaking/types/pop.go#L200

x/btcstaking/types/pop.go
RafilxTenfen: done at https://github.com/babylonchain/babylon/pull/657/commits/dbdffc6e5328bfa99935507f44a61018e5f4ae7a

x/btcstaking/keeper/msg_server.go
RafilxTenfen: It does check for us because otherwise it cannot validate the signing, since it is the signer 
But I also added to the [ValidateBasic](https://github.com/babylonchain/babylon/blob/280d8868200ddd742bbe78e97942909fcd2aeb22/x/btcstaking/types/msg.go#L81) 

on the msg server is needed to parse from `string` to `sdk.AccAddress` due to other functions expect to receive that type instead of `string`

KonradStaniec (approved):
LGTM!

gitferry (approved):
Nice work!

KonradStaniec (approved):
LGTM! Great work pushing this forward!

PR #660, opened 2024-06-03
chore: remove pop of `BTCDelegation` and emit `EventNewBTCDelegation`

RafilxTenfen:
- Reduce storage of BTCDelegation
- Add event for other processes that want to retrieve and make sure the Proof Of Possession matches


proto/babylon/btcstaking/v1/events.proto
```diff
@@ -48,3 +49,13 @@ message EventPowerDistUpdate {
...
+
+// EventNewBTCDelegation is the event emitted when an new
+// BTC delegation is created.
+message EventNewBTCDelegation {
+  // delegation is the new BTC delegation created.
+  BTCDelegation delegation = 1;
+  // pop identifies the Proof of Possession of the BTC by the
+  // staker address.
+  ProofOfPossessionBTC pop = 2;
+}
```
SebastianElvis, line 61: Tbh I prefer keeping PoP in DB as

- We already have event `EventBTCDelegationStateUpdate` for capturing state updates of BTC delegations, including the newly pending ones. The extra thing in this event is only PoP which is a bit marginal
- The event cannot be subscribed by the consumer chains which need to receive and verify PoP from Babylon.
- PoP is not the storage overhead hotspot and we might not get much improvement on the storage cost by removing it from DB

RafilxTenfen: IMO if it is not needed for anything else after the delegation is created, it should be removed
Kindly reminder that our bbn-test-3 reached 2TB storage in a few months, even if we reduce just a few GB it is already something 

@SebastianElvis  Can I ask why consumer chains need to verify PoP if it was already verified by bbn?

SebastianElvis: It's because consumer chains might not trust Babylon

RafilxTenfen: if they do not trust Babylon they would need to stop using the BTC finality round entirely, correct?

SebastianElvis: This depends. Consumer does not trust Babylon in the sense that it will verify everything coming from Babylon. It will stop using Babylon only when it finds Babylon is indeed behaving adversarial, e.g., sending invalid IBC packets to consumers or censoring certain IBC packets.

RafilxTenfen: Understood, it makes sense to leave PoP for them to verify as well...

PR #661, opened 2024-06-04
chore: Merge dev to base/jail-inactive-finality-provider

gitferry:
This PR is to rebase to dev before going forward on [base/jail-inactive-finality-providers](https://github.com/babylonchain/babylon/tree/base/jail-inactive-finality-providers)


SebastianElvis (approved):
Probably you could simply do a rebase as the target branch is a base branch rather than dev after all

Issue #666, opened 2024-06-07
chore: add gosec


RafilxTenfen:
https://github.com/securego/gosec

PR #667, opened 2024-06-07
chore: add gosec

RafilxTenfen:
- Fix some errors checks
- Add gosec to CI
- Add gosec to makefile 
  - `gosec` runs in docker
  - `gosec-local` runs for local gosec assuming it is available in $PATH

Closes: #666


Makefile
```diff
@@ -340,6 +341,18 @@ format:
...
+###############################################################################
+###                                Gosec                                    ###
+###############################################################################
+
+gosec:
+	$(DOCKER) run --rm -it -w /$(PROJECT_NAME)/ -v $(CURDIR):/$(PROJECT_NAME) securego/gosec -exclude=G404 -exclude-generated -exclude-dir=/$(PROJECT_NAME)/testutil -exclude-dir=/$(PROJECT_NAME)/test -conf /$(PROJECT_NAME)/gosec.json /$(PROJECT_NAME)/...
+
+gosec-local:
+	gosec -exclude=G404 -exclude-generated -exclude-dir=$(CURDIR)/testutil -exclude-dir=$(CURDIR)/test  -conf $(CURDIR)/gosec.json $(CURDIR)/...
+
```
SebastianElvis, line 353: Why do we exclude G404 sec rule?

x/btclightclient/keeper/keeper.go
```diff
@@ -47,10 +48,21 @@ func NewKeeper(
 	}
 }
 
-func (k Keeper) Logger(ctx sdk.Context) log.Logger {
+func Logger(ctx sdk.Context) log.Logger {
 	return ctx.Logger().With("module", fmt.Sprintf("x/%s", types.ModuleName))
 }
 
+// EmitTypedEventWithLog emits an event and logs if it errors.
+func EmitTypedEventWithLog(ctx context.Context, evt proto.Message) {
```
SebastianElvis, line 56: Can we 

- Tie these two functions under `Keeper`?
- Make `EmitTypedEventWithLog` private as it's not supposed to be called by external callers?

SebastianElvis (approved):
That's a great move!

Makefile
RafilxTenfen: G404 does not let us use `math/rand` as import, but some functions on the cosmos-sdk itself ask for that 
In our case, mostly in the `app/test_helpers.go` 

I could also exclude this file from the gosec search if you prefer than exclude the G404 rule

x/btclightclient/keeper/keeper.go
RafilxTenfen: I agree to make it private, but I do not see a reason to tie to the keeper if they don't use any information from the keeper itself

RafilxTenfen: I can move these two functions to `x/btclightclient/keeper/utils.go`

RafilxTenfen: https://github.com/babylonchain/babylon/pull/667/commits/b2e027e0bb2b1e7c631c63bd7f939ba9c65bbc81

Makefile
SebastianElvis: If those are test helpers then probably excluding the files is better? G404 is about checking the usage of vulnerable crypto primitives so imo is a meaningful check

x/btclightclient/keeper/keeper.go
SebastianElvis: > I agree to make it private, but I do not see a reason to tie to the keeper if they don't use any information from the keeper itself

Not a strong requirement, but rather a convention -- you could see other keepers have `Logger` function. No strong objection though

Makefile
RafilxTenfen: agreed

x/btclightclient/keeper/keeper.go
RafilxTenfen: made sense, done at https://github.com/babylonchain/babylon/pull/667/commits/41a955a3ba19d5411f0628619181f034e7ea32b9

Makefile
RafilxTenfen: turns out we were not even using that func '-' 
Removed at https://github.com/babylonchain/babylon/pull/667/commits/368df2e40e2fd5bc9f591db9ed708ef7498d31b8

PR #670, opened 2024-06-11
chore: add test for btc multisig feegrant

RafilxTenfen:
- Add new Tx command to create FeeGrant
- Test to create BTC delegation with staker out of BBN funds and granter paying for the fees


gitferry (approved):
LGTM! Nit: maybe we should rebase the base branch, ie, `rafilx/btc-multisig` other than do it here, so that we don't need to review the changes in `dev`

PR #675, opened 2024-06-17
DNM: Command to print size and number of node for every Babylon module

KonradStaniec:
just for testing/deployment purposes


cmd/babylond/cmd/ModuleSizesCmd.go
RafilxTenfen, line 1: to be honest this would actually be a nice command to have '-'

What do you think to put it under `babylond debug` ?

KonradStaniec:
closed in favour of - https://github.com/babylonchain/babylon/pull/687

PR #678, opened 2024-06-24
chore: FP as signer addr

RafilxTenfen:
- [ADR 021](https://github.com/babylonchain/pm/blob/d24b819fd8a0211d6888bcb4a640d01862c85dd7/adr/adr-021-fp-signer.md)

TODO: 
- Update finality-provider
- Test with deployments
- Add/update unit tests


x/btcstaking/keeper/msg_server.go
```diff
@@ -116,15 +122,20 @@ func (ms msgServer) EditFinalityProvider(ctx context.Context, req *types.MsgEdit
 		return nil, types.ErrCommissionGTMaxRate
 	}
 
+	// TODO: check to index the finality provider by his address instead of the BTC pk
```
RafilxTenfen, line 125: @SebastianElvis 

what is your take on this? To index or key the kv store by the FP addr instead of the BTC PK
I raised this question also because of editing, someone can edit their BTC pk, but not their bbn address

SebastianElvis: Given that we now have BTC staking integration and consumer chains won't have Babylon addresses, let's stick to BTC PK for the key then

SebastianElvis (approved):
Finallit got rid of the full PoP thing. great work!

x/btcstaking/keeper/msg_server.go
RafilxTenfen: got it, makes sense
Thanks!

x/btcstaking/keeper/msg_server.go
```diff
@@ -116,15 +122,20 @@ func (ms msgServer) EditFinalityProvider(ctx context.Context, req *types.MsgEdit
...
 
+	fpAddr, err := sdk.AccAddressFromBech32(req.Addr)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid address %s: %v", req.Addr, err)
+	}
+
 	// ensure the signer corresponds to the finality provider's Babylon address
-	fpBabylonAddr := sdk.AccAddress(fp.BabylonPk.Address())
-	if req.Signer != fpBabylonAddr.String() {
+	if !strings.EqualFold(fpAddr.String(), fp.Addr) {
```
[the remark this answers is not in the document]
RafilxTenfen: nop, sticking with index by BTC pk

gitferry (approved):
Great work! It would be good to also update `x/btcstaking/README.md`

RafilxTenfen:
> Great work! It would be good to also update `x/btcstaking/README.md`

Good catch, done at https://github.com/babylonchain/babylon/pull/678/commits/1f0d07516c8b3d74ba9bb291409dc273672de286

KonradStaniec (approved):
Great job pushing this forward! 👍

client/client/tx.go
RafilxTenfen, line 1: @SebastianElvis @KonradStaniec @gitferry 
You guys already approved the PR, but I had modified this file to relay msgs using generated private key, it was useful for e2e test https://github.com/babylonchain/btc-staker/pull/178/files#diff-d7a3bcca6ff45d1e15c3e310bfd93ce0c074f276a184d7c8c38445ee32826b7dR697

PR #679, opened 2024-06-25
Enforce that keys are unique in scripts

KonradStaniec:
Enforce that `StakerPK ` is never equal to `FinalityProviderPk` as this breaks our security with pre-signed slashing transactions.

If `StakerPK == FinalityProviderPk` in script, and Staker publishes pre-signed slashing transaction, then quorum of covenant committee  member can collude to slash the staker by creating valid witness in which staker signature will be used in place of finality provider signature.


docs/staking-script.md
```diff
@@ -173,12 +181,12 @@ The timelock path locks the staker's Bitcoin for a pre-determined number of
 Bitcoin blocks. It commits to a script of the form:
 
 ```
-<Staker_PK> OP_CHECKSIGVERIFY  <Timelock_Blocks> OP_CHECKSEQUENCEVERIFY
+<StakerPK> OP_CHECKSIGVERIFY  <Timelock_Blocks> OP_CHECKSEQUENCEVERIFY
```
gitferry, line 184:
```suggestion
<StakerPK> OP_CHECKSIGVERIFY  <TimelockBlocks> OP_CHECKSEQUENCEVERIFY
```

gitferry (approved):
Nice fix!

SebastianElvis:
Agree that having the same Pk for staker / FP will incur this new attack vector launched by covenant committee. But in this case how will we support self delegation?

KonradStaniec:
> Agree that having the same Pk for staker / FP will incur this new attack vector launched by covenant committee. But in this case how will we support self delegation?

I think we do not have defined how self-delegation will be supported.

In our pm epic https://github.com/babylonchain/pm/issues/41, we have this  quick note in nice to have:
```
5. Self-Delegation of Finality Providers
Why nice to have: currently it is possible to self delegate but it requires jumping through some hops and there is no easy way to prove that delegation is really self delegation. So this is bad but not critical to launch initial version of mainnet.
```

There is even a question, does the Babylon node/protocol need to support native self-delegation ? As I see it, self delegation is made when some entity control both `StakerPK` and `FinalityProviderPK`, this means it can be actually supported off chain . For example, some front end which receive proof of possession for both `StakerPK` and `FinalityProviderPK`.

If this is not enough and we would like to support it natively, maybe the option would be the have additional message for finality provider like `ProveSelfDelegationMsg` in which finality provider send transactions hashes which were self delegated.

Either way, we cannot do it by setting `StakerPk == FinalityProviderPk` as this is not secure.

docs/staking-script.md
```diff
@@ -149,6 +149,14 @@ The requirements for a valid slashing transaction are:
 
 ## Staking and Unbonding output scripts
 
+In the below scripts, there are three entities, each represented by BTC public
+key:
+- `StakerPK` - BTC staker public key
```
SebastianElvis, line 154:
```suggestion
key:

- `StakerPK` - BTC staker public key
```

btcstaking/types.go
```diff
@@ -278,6 +290,14 @@ func newBabylonScriptPaths(
 		return nil, fmt.Errorf("staker key is nil")
 	}
 
+	if ContainsKey(fpKeys, stakerKey) {
+		return nil, fmt.Errorf("finality provider keys contain staker key")
+	}
+
+	if ContainsKey(covenantKeys, stakerKey) {
+		return nil, fmt.Errorf("covenant keys contain staker key")
+	}
```
SebastianElvis, line 299: Should we also check `fpKeys` and `covenantKeys` do not overlap? If they overlap then a slashed FP will compromise a covenant member, which is certainly not good.

Then a better way to implement this might be creating a new array containing all these keys, sort them, and check dup

KonradStaniec: agree on check, better to be safe here.

As for implementation, probably using local map to check for duplicates is a bit cleaner, but I will play with both.

KonradStaniec:
@gitferry @SebastianElvis In last commit:
- I have added check for global uniqness
- Added new sentinel public error `ErrDuplicatedKeyInScript` which can be used by caller to discern this error. Imo some CLI tools can add more user friendly error message based on this error
- removed uniqness check from `prepareKeysForMultisigScript` and added in doc there that ` It is up to the caller to ensure that the keys are unique` . This function is private so it makes sense to move uniqness check higher up,  and not do double checking.

btcstaking/types.go
```diff
@@ -267,6 +270,38 @@ type babylonScriptPaths struct {
...
+func keyToString(key *btcec.PublicKey) string {
+	return hex.EncodeToString(schnorr.SerializePubKey(key))
+}
+
+func checkForDuplicateKeys(
+	stakerKey *btcec.PublicKey,
+	fpKeys []*btcec.PublicKey,
+	covenantKeys []*btcec.PublicKey,
+) error {
+	keyMap := make(map[string]bool)
```
SebastianElvis, line 282:
```suggestion
	keyMap := make(map[string]struct{})
```

is better, see https://www.reddit.com/r/golang/comments/qzha6o/for_sets_maptstruct_vs_maptbool/

SebastianElvis (approved):
Nice fix!

btcstaking/types.go
KonradStaniec: it is better only if you care about memory ie the map is predicted to be large, here I would say it is not the case and imo version with bools is a bit more readable.

Saying all that, it is not that strong opinion, so I will change it 👍

PR #680, opened 2024-06-25
Tighten checks in CheckTransactions function

KonradStaniec:
- make `validateSlashingTx` private as it should not be part of the public api, the validation of slashing tx only have point in context of validation of funding/slashing transactions pair
- add check for slashing and funding transaction -  `blockchain.CheckTransactionSanity` which check transactions against BTC rules which are relevant to overlfows and transactions limits.

### DIFF
--- a/btcstaking/staking.go
+++ b/btcstaking/staking.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 
 	sdkmath "cosmossdk.io/math"
+	"github.com/btcsuite/btcd/blockchain"
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcec/v2/schnorr"
 	"github.com/btcsuite/btcd/btcutil"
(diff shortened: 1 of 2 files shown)


btcstaking/staking.go
```diff
@@ -344,6 +346,18 @@ func CheckTransactions(
 	slashingChangeLockTime uint16,
 	net *chaincfg.Params,
 ) error {
+	if slashingTx == nil || fundingTransaction == nil {
+		return fmt.Errorf("slashing and funding transactions must not be nil")
+	}
+
+	if blockchain.CheckTransactionSanity(btcutil.NewTx(slashingTx)) != nil {
+		return fmt.Errorf("slashing transaction does not obey BTC rules")
```
gitferry, line 354: maybe we can wrap the error?

btcstaking/staking.go
```diff
@@ -344,6 +346,18 @@ func CheckTransactions(
...
+	}
+
+	if blockchain.CheckTransactionSanity(btcutil.NewTx(fundingTransaction)) != nil {
+		return fmt.Errorf("funding transaction does not obey BTC rules")
```
gitferry, line 358: same here

gitferry (approved):
Nice fix! A nit:

RafilxTenfen (approved):
LGTM

PR #681, opened 2024-06-27
feat(jailing): Identify inactivity upon EndBlocker

gitferry:
This PR (part of #682) is to impelement liveness handler within `EndBlocker`. Some notable changes:

* Implement algorithm to update the signing info and detect inactive finality providers.
* Add hook `AfterFinalityProviderActivated` to `x/btcstaking` and implementation in `x/finality` to create a signing info when a finality provider is activated. By `activated`, we mean the finality provider enters the active set for the first time or re-enters the set
* Add `inactive` flag to the finality provider to indecate whether inactivity of the finality provider is detected
* Add hook `AfterInactiveFinalityProviderDetected` to `x/finality` and implementation in `x/btcstaking` to set the `inactive` flag of the finality provider to `true`
* Fuzz test of livenss handler


x/finality/keeper/hooks.go
```diff
@@ -0,0 +1,43 @@
...
+func (h Hooks) AfterFinalityProviderActivated(ctx context.Context, fpPk *bbntypes.BIP340PubKey) error {
+	signingInfo, err := h.k.FinalityProviderSigningTracker.Get(ctx, fpPk.MustMarshal())
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	if err == nil {
+		signingInfo.StartHeight = sdkCtx.BlockHeight()
+	} else if errors.Is(err, collections.ErrNotFound) {
+		signingInfo = types.NewFinalityProviderSigningInfo(
+			fpPk,
+			sdkCtx.BlockHeight(),
+			time.Unix(0, 0),
```
RafilxTenfen, line 37: can we use last block time? 
not good feelings about having some kind of mismatch by using current time of server

RafilxTenfen: ohh, this is just temporary and will be defined latter for how long the fp is jailed, got It

x/finality/keeper/signing_info.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// SetMissedBlockBitmapValue sets, i.e. flips, a bit in the validator's missed
+// block bitmap. When missed=true, the bit is set, otherwise it set to zero. The
+// index provided is assumed to be the index in the range [0, SignedBlocksWindow),
+// which represents the bitmap where each bit represents a height, and is
+// determined by the validator's IndexOffset modulo SignedBlocksWindow. This
+// index is used to fetch the chunk in the bitmap and the relative bit in that
+// chunk.
+func (k Keeper) SetMissedBlockBitmapValue(ctx context.Context, fpPk *bbntypes.BIP340PubKey, index int64, missed bool) error {
```
RafilxTenfen, line 51: can we add hardcoded tests for the bitmap, it could also be done in another PR...

x/finality/keeper/hooks.go
RafilxTenfen: it is correctly updated at https://github.com/babylonchain/babylon/pull/681/files#diff-df98eef61c127aae58fcd4a524f64979aa40e04772f202736881c9b9927a4b14R147

Makes sense

RafilxTenfen (approved):
LGTM! :tada:

x/finality/keeper/signing_info.go
gitferry: Hardcoded test function or hardcoded parameters?

x/finality/keeper/liveness.go
```diff
@@ -2,10 +2,180 @@ package keeper
...
+
+func (k Keeper) HandleFinalityProviderLiveness(ctx context.Context, fpPk *types.BIP340PubKey, missed bool, height int64) error {
+	params := k.GetParams(ctx)
+	// don't update missed blocks when finality provider is already jailed
+	val, err := k.BTCStakingKeeper.GetFinalityProvider(ctx, fpPk.MustMarshal())
+	if err != nil {
+		return err
+	}
+
+	if val.IsJailed() {
```
KonradStaniec, line 47: shouldn't we also check if finlity provider is not slashed ? or such providers won't be retrieved in that function ?

KonradStaniec (commented):
some initial high level comments/questions to better understand the flow here.

x/btcstaking/keeper/finality_providers.go
```diff
@@ -67,6 +67,28 @@ func (k Keeper) SlashFinalityProvider(ctx context.Context, fpBTCPK []byte) error
...
+// JailFinalityProvider jails a finality provider with the given PK
+func (k Keeper) JailFinalityProvider(ctx context.Context, fpBTCPK []byte) error {
+	// ensure finality provider exists
+	fp, err := k.GetFinalityProvider(ctx, fpBTCPK)
+	if err != nil {
+		return err
+	}
+
+	// ensure finality provider is not slashed yet
+	if fp.IsJailed() {
```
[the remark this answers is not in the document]
gitferry: Good point. We should not jail slashed finality providers as their voting power is already removed

x/finality/types/signing_info.go
```diff
@@ -0,0 +1,20 @@
...
+
+// NewFinalityProviderSigningInfo creates a new FinalityProviderSigningInfo instance
+func NewFinalityProviderSigningInfo(
+	fpPk *bbntypes.BIP340PubKey, startHeight int64,
+	jailedUntil time.Time, missedBlocksCounter int64,
+) FinalityProviderSigningInfo {
+	return FinalityProviderSigningInfo{
+		FpBtcPk:             fpPk,
+		StartHeight:         startHeight,
+		JailedUntil:         jailedUntil,
```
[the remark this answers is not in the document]
gitferry: I agree that using `block_height` is more intuitive in blockchain world but here is following the Cosmos-sdk's [convention](https://github.com/cosmos/cosmos-sdk/blob/562723cfa9c824ec2051e3d92d1f0c6aedaf39a1/x/slashing/types/slashing.pb.go#L45). I think it's better to follow this convention as most of our fp operators are familiar with running cosmos nodes unless there's any security issue

x/finality/keeper/liveness.go
gitferry: We should return here if the fp is slashed as we don't remove slashed fps from db

x/finality/abci.go
```diff
@@ -27,7 +28,9 @@ func EndBlocker(ctx context.Context, k keeper.Keeper) ([]abci.ValidatorUpdate, e
 		// tally all non-finalised blocks
 		k.TallyBlocks(ctx)
 		// jail inactive finality providers if there are any
-		k.HandleLiveness(ctx)
+		// TODO: decide which height to use the handle liveness
```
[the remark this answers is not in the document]
gitferry: We do allow finality signatures submitted after block finalized so adding a parameter like `finality_timeout` sounds good to me and that's also the point why there's a TODO here. Will do it in a separate PR

> I wonder whether `EndBlocker` is place to do it.
Are you suggesting we put it in `BeginBlocker`? This is what Cosmos-sdk does but not sure we have to put it there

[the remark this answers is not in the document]
KonradStaniec: I am thinking about having hook `AfterBabylonBlockFinalized` , and doing this liveness operations when this hook is called, as from liveness detector pov this is is most important event.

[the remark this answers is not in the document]
gitferry: what if no block is finalized? If the finality is stuck then we cannot handle inactive finality providers, can we?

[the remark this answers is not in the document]
KonradStaniec: good point. So we have conceptually two different cases:
1. blocks are getting finalized, but some of the fps did not vote
2. finality is stalled, so host chain progresses but finality is behind

`1` can be handled by `AfterBabylonBlockFinalized`
For `2` we would need another hook `AfterStallingDetected`

From what I understand `HandleLiveness` tries to handle both cases now, which makes it a bit complex. Maybe making explicit that those two cases exist, would make things clearer ?

x/finality/keeper/liveness.go
```diff
@@ -2,10 +2,180 @@ package keeper
...
+		missed := !ok
+
+		err = k.HandleFinalityProviderLiveness(ctx, fpPk, missed, height)
+		if err != nil {
+			panic(fmt.Errorf("failed to handle liveness of finality provider %s: %w", fpPkHex, err))
+		}
+	}
+}
+
+func (k Keeper) HandleFinalityProviderLiveness(ctx context.Context, fpPk *types.BIP340PubKey, missed bool, height int64) error {
```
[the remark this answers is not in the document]
gitferry: Some refactoring done in [66e7959](https://github.com/babylonchain/babylon/pull/681/commits/66e79595473edf16151e9cc736cb91a849eee8fb)

x/finality/abci.go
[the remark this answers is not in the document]
gitferry: For detecting stalling, don't we also need a parameter `finality_timeout`? So I don't see distinguishing the two cases would make things simpler. I think nevertheless, we need to introduce a timeout parameter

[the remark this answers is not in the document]
KonradStaniec: yup, this is also something I am sure about `finality_timeout` will be needed 👍 

About distinguishing the two cases explicitly, for now this is not a blocker we can re-visit it later after adding `finality_timeout`.

x/btcstaking/keeper/finality_providers.go
```diff
@@ -67,6 +67,28 @@ func (k Keeper) SlashFinalityProvider(ctx context.Context, fpBTCPK []byte) error
...
+// JailFinalityProvider jails a finality provider with the given PK
+func (k Keeper) JailFinalityProvider(ctx context.Context, fpBTCPK []byte) error {
+	// ensure finality provider exists
+	fp, err := k.GetFinalityProvider(ctx, fpBTCPK)
+	if err != nil {
+		return err
+	}
+
+	// ensure finality provider is not slashed yet
+	if fp.IsJailed() || fp.IsSlashed() {
```
KonradStaniec, line 79: Two more questions

x/btcstaking/keeper/power_dist_change.go
```diff
@@ -54,6 +56,14 @@ func (k Keeper) UpdatePowerDist(ctx context.Context) {
 	// to construct the new distribution
 	newDc := k.ProcessAllPowerDistUpdateEvents(ctx, dc, events, maxActiveFps)
 
+	// find newly bonded finality providers and execute the hooks
+	newBondedFinalityProviders := newDc.FindNewActiveFinalityProviders(dc, maxActiveFps)
```
KonradStaniec, line 60: two more questions:
1. Suppose in last block we jailed some finality provider, should he be in in the next `VotingPowerDistCache` ? For now I see he will be. (Unless I am missing something)
2. A bit tangent, when receiving votes on finality, we reject votes from slashed fps.
```	
if fp.IsSlashed() {
return nil, bstypes.ErrFpAlreadySlashed
}
```
shouldn;t we do the same for jailed ones ?

x/btcstaking/keeper/keeper.go
```diff
@@ -22,6 +23,8 @@ type (
 		btccKeeper  types.BtcCheckpointKeeper
 		ckptKeeper  types.CheckpointingKeeper
 
+		hooks types.BtcStakingHooks
```
SebastianElvis, line 26: I remember `hooks` object is constructed upon calling `k.Hooks()`. Why does it need to be a part of the keeper here?

x/finality/abci.go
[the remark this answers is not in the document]
SebastianElvis: > finality is stalled, so host chain progresses but finality is behind

I'm not sure jailing inactive FP can help with this case. Suppose the finalised blocks are stalled and some FPs are jailed due to the inactivity across these blocks. Then

- If we allow these jailed FPs to vote for these blocks, then they don't have incentive to do so as they are jailed anyway. Also this has some implication on the implementation -- the chain needs to remember the height when the FP is jailed.
- If we don't allow these jailed FPs to vote for these blocks, then these blocks cannot be finalised forever.

Have we decided to jail FPs when finality is stalled? from incentive pov it seems that we need to encourage FPs to vote (eg forget jailing inactivity a bit) when finality is stalled

x/btcstaking/keeper/power_dist_change.go
KonradStaniec: > Should it be part of story 2? The voting power should be removed first, then we reject votes

Just thinking about it, maybe story 1 should end up with just implementing hook/callback `AfterDectingFinalityProviderInactivity(fpPK)` without actually jailing ? At the beginning we can just log it, and later implement jailing ?

x/finality/abci.go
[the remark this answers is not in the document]
gitferry: > If we don't allow these jailed FPs to vote for these blocks, then these blocks cannot be finalised forever.

Jailing will remove voting power of jailed fps from the total voting power, so that the quorum will be easier to be reached

x/btcstaking/keeper/power_dist_change.go
gitferry: In my mind for story 1, inactive finality providers will be marked as `jailed` without actuall punishment. Users can query the `jailing` status of fps externally and can `unjail` them after jailing period is passed by sending `Unjail` message. I think this sets a more solid base for the next step of apply actual punishment. Only logging seems opaque, wdyt?

x/btcstaking/keeper/keeper.go
gitferry: I think what you described is for keepers that implement the hooks. For keepers that call the hooks, they need to have hooks as part of the keeper. See [x/btclightclient](https://github.com/babylonchain/babylon/blob/58c28046b75d3e1e6bd014ed34d7bea75665bf80/x/btclightclient/keeper/keeper.go#L24) and [x/staking](https://github.com/cosmos/cosmos-sdk/blob/3c13bc03cdb12f38c3a7aa66995df663c360e17d/x/staking/keeper/keeper.go#L76)

x/btcstaking/keeper/power_dist_change.go
KonradStaniec: Hmm in general this reasoning seems sound. 

I think my biggest problem with it is that `Jailed` is loaded and known term and everybody looking at codebase will assume:
- that if somebody is jailed it means he do no have voting power
- jaling is actual punishment

Form point of view of story 1 what we do, is marking a fp `Inactive`, it is not a punishment but rather status change.  I think if we go without actual punishment then naming shoud be changes at least. (maybe `active/inactive` or somehing similiar?)

gitferry: In prod we would have both story 1 and story 2, right? We can use `Inactive` for now, but we will eventually change it to `Jailed`, won't we? Why bother?

KonradStaniec: we do not know when will story 2 get implemented/finished, we also do not know particular details of our jailing mechanism, also we will have new core team members, having the `Jailing` name without jailing implementation will be confusing.

I for one, was confused when looking at this pr and seeing jailing. 

Wdyt about:
- having hook `AfterFinalityProviderInactivityDeteced`
- having boolean flag `Active` , as I totally agree with you that having such status is nice for many purposes
- in case of inactivity `AfterFinalityProviderInactivityDeteced` hook sets flag to `false`
- in case of first fp vote after inactivity was detected, flag is set to `true`

This way:
- we are explicit that we are only tracking activity of fp, without any additional punishments, and we can test this mechanism separately.
- later when implementing jailing, most of the changes will be additive ie we will remove `Active` flag and and `Jailed` flag, remove voting power etc. The implementation of activity tracking (`AfterFinalityProviderInactivityDeteced` hook) will remain the same.

I know this adds a bit of overhead in whole impl (story 1 and story 2), but makes both stories totally self contained and separate.

gitferry: In general, I agree with the idea that we should keep story 1 and story 2 self-contained and separate.
* Agreed that we should have a hook `AfterFinalityProviderInactivityDeteced`. For story 1, `x/btcstaking` will implement it to set the flag. For story 2, `x/incentive` might implement it. Done in [a8fe571](https://github.com/babylonchain/babylon/pull/681/commits/a8fe5712c9cdc7e774dff69a6230649bdb1a167b)
* Not sure we should have `Active` flag because it has mixed meaning of active set. I think here we should have `Inactive` to share the same reaction we would do for `jailing`. The value will be `false` as default, and set to `true` upon `AfterFinalityProviderInactivityDeteced`. See [1e303a3](https://github.com/babylonchain/babylon/pull/681/commits/1e303a3908ae85508becf2164964175d9d4c455a)
* For story 1, we probably don't need the concept of jailing period, so I removed it in [1e303a3](https://github.com/babylonchain/babylon/pull/681/commits/1e303a3908ae85508becf2164964175d9d4c455a)
* In a subsequent PR, will implement a msg like `MsgRecoverFinalityProvider` (corresponding to `MsgUnjailFinalityProvider`) to set flag `Inactive` to be `false`

x/finality/keeper/signing_info.go
RafilxTenfen: Hardcoded test function

KonradStaniec (approved):
looks good! thanks for accomodating to the comments 👍 

One note though, given that inactivity is purely informational flag (without any penalies) which will probably be triggered when fp misses his vote, maybe it makes sense to bring this flag to `false` after next successful vote instead of having separate message ?  This can be done in separate pr though.

x/btcstaking/keeper/keeper.go
SebastianElvis: Not necessarily, it can be created on-demand, see https://github.com/babylonchain/babylon/blob/416912aa80d1b4de9539ef17772f353edf9a2f7c/x/epoching/keeper/hooks.go#L47-L48

I feel creating hook object on demand is slightly preferable as the hook is more like a "client" that invokes the keeper rather than a part of the state of the keeper. No strong opinion on this, your call

x/finality/abci.go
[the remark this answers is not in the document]
SebastianElvis: > Jailing will remove voting power of jailed fps from the total voting power, so that the quorum will be easier to be reached

This is true for future blocks, but not for blocks that are already generated but haven't got enough finality signatures, right? as we won't modify voting power table of previous blocks

x/finality/keeper/liveness.go
```diff
@@ -2,10 +2,183 @@ package keeper
...
+		err = k.hooks.AfterInactiveFinalityProviderDetected(ctx, fpPk)
+		if err != nil {
+			return err
+		}
+
+		// We need to reset the counter & bitmap so that the finality provider won't be
+		// immediately detected as inactive upon re-bonding.
+		// We don't set the start height as this will get correctly set
+		// once they bond again in the AfterFinalityProviderActivated hook
+		signInfo.MissedBlocksCounter = 0
```
SebastianElvis, line 94: A number of operations could be wrapped as functions of `FinalityProviderSigningInfo`, e.g., incrementing/decrementing/resetting the counter, checking whether inactivity is detected, etc.. this is not a blocker for this PR though

SebastianElvis (approved):
Lgtm 👍 completely agree with the discussion that this PR focuses on detecting inactive FPs upon `EndBlock`

gitferry:
> One note though, given that inactivity is purely informational flag (without any penalies) which will probably be triggered when fp misses his vote, maybe it makes sense to bring this flag to false after next successful vote instead of having separate message ? This can be done in separate pr though.

@KonradStaniec Good point. For story 1, we don't need to send a separate message to set the `inactive` to `false`. But I don't think we should set it to `false` upon next successful vote. It should also depend on the same detection algorithm. Specifically, we can set it back to `false` if the missed block counter is below the threshold and the `inactive` was set to `true`.

x/finality/abci.go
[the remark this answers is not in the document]
gitferry: I was thinking if finality halting is detected (need a new parameter for this), we would enter an emergency mode where we need to modify voting power table of all the non-finalized blocks by removing jailed voting power and retry finalization. Cosmos does not need to worry about this because the jailing is done in the BeginBlocker of the next block

x/btcstaking/keeper/keeper.go
gitferry: `x/epoching` needs to implement hooks from from `x/staking` and `x/checkpointing`, so `func (k Keeper) Hooks() Hooks { return Hooks{k} } ` is needed for that. Moreover, other modules need to implement hooks of `x/epoching`, so its keeper also has [hooks](https://github.com/babylonchain/babylon/blob/416912aa80d1b4de9539ef17772f353edf9a2f7c/x/epoching/keeper/keeper.go#L18)

PR #683, opened 2024-06-28
typo: Update btcstaking_test.go

puneet2019:
Name finalityProviderKeys as finalityProvider keys instead of covenant privkeys.  Let's not confuse people here even if key types are same..


puneet2019:
@RafilxTenfen any reason for this CI to fail? doesn't seem related to changes...

RafilxTenfen:
Humm, not sure also it should not fail :thinking:

PR #682, opened 2024-06-27
feat(inactivity): Identify inactive finality providers

gitferry:
This PR is to collect small PRs to close #613 

Tasks:

- [x] Add finality provider signing info tracker and params, #614
- [x] Identify inactivity upon EndBlocker, #681
- [x] Add timeout to params and logic for reverting inactive flag, [context](https://github.com/babylonchain/babylon/pull/681#discussion_r1658290121), #689

PR #689, opened 2024-07-04
feat(Inactivity): Add LivenessDelay to finality params and logic for reverting inactive flag

gitferry:
This PR is part of #682 and a follow-up to https://github.com/babylonchain/babylon/pull/681#discussion_r1658290121. Specifically, this PR:
* Add `LivenessDelay` to `x/finality` params, which sepcifies a number of heights after which we examine the liveness of finality providers at a certain height. For example, the current height is `100` and `params.LivenessDelay = 3`, then we will deal with liveness issues at height `97`. This is introduced due to the fact that we allow finality sigs after quorum.
* We revert the `inactive` flag of a finality provider according to the same algorithm that setting the `inactive` flag to `true`. For example, in a signed window of `100`, we determine a fp to be `inactive` if it misses more than 50 blocks. So, if this fp is already `inactive`, it will be reverted at the height where the missed count is reduced to `50` in the window.


proto/babylon/finality/v1/params.proto
```diff
@@ -12,9 +12,11 @@ message Params {
   option (gogoproto.goproto_stringer) = false;
   // signed_blocks_window defines the size of the sliding window for tracking finality provider liveness
   int64 signed_blocks_window  = 1;
+  // liveness_delay defines the delay for examining inactivity of finality providers
+  int64 liveness_delay = 2;
```
SebastianElvis, line 16: Does `liveness_delay` have some overlap with `min_signed_per_window` in some sense? Say we enforce an FP to sign >=50 blocks in the last 100 blocks. At height 100, the `liveness_delay` is 100 for block 1, 99 for block 2, etc.. 

If this is true, does that mean we don't need the `liveness_delay` at all?

Also, the name might be improved. The parameter defines the maximum delay of a finality sig on a block, but the name suggests it has something to do with the liveness property

gitferry: I'm not sure why it is related to `min_signed_per_window`, but this is a delay to allow more finality sigs to be collected. In cometBFT, liveness is checked upon next block because sigs will not be collected once a quorum is reached. However, we don't have this rule, so that we can wait for a few blocks to examine the liveness.

As for naming, I agree that this is not directly related to liveness. Perhaps `inactivity_check_delay` or just `inactivity_delay`?

SebastianElvis: I was saying that `min_signed_per_window` seems to already allow some delay for submitting finality signatures. Suppose the chain is checking the window `[x-100, x]`, FP has 100 block time for voting block `x-100`, and 99 block time for voting block `x-99`, etc... If `min_signed_per_window` is small, then FP can take more time to submit finality signatures. Or do I miss something?

gitferry: The total missed blocks of a window is updated per block. For example, `min_signed_per_window = 50` in a window size of 100, a fp has missed the second 50 blocks of the window. It will be recognized as inacitve if it misses the next block

KonradStaniec: I think the problem we have here is that we are conflating at least two issues here. Whole inactivity detection mechanism have two problems two solve:
1. Detection of liveness issues of each particular finality provider
2. Detection of liveness issues of whole finality mechanism

To solve issue `1`, all of the calculations needs to take place relative to last finalized block and not relative to the tip of the babylon chain. So if we have this `timeout_parameter` (or `liveness_delay` as it is called in this pr) , it should takes place when block are getting finalized. Example:
1. we have `liveness_delay = 1`
2. when block `n` is finalised,  we should consider block `n-1` as not accepting more votes, and calculate inactivity statistics for all of the finality providers
From what I understand this is how cometbft is working currently ie it accepts votes after the quorum is met but only until next block is committed.

To solve issue `2`, we will need to compare the difference `current_best_babylon_block - last_finalized_block`  with some threshold, and if difference is larger that this threshold then we will need to do something.

Does this thinking goes in line with this pr ?

gitferry: I see your point but I'm not convinced that we should detect liveness upon a block is finalized. My argument is that:
1. This forces us to consider two scenarios like you described, happy case and finality halting, and have different algorithms to the two scenarios. We also need to consider the switch of the two cases.
2. To deal the finality halting case, it seems that we need to fall back to the mechanism that this pr implements: calculations relative to the tip of the babylon chain
3. Finalization already takes a lot of computation efforts (distribute reward etc.). If we trigger liveness detection also based on finalization, it might make finalization heavier

In the proposed approach, all the calculations take place relative the the tip of the babylon chain with the delay of `liveness_delay`, which I think already resolves problem `1` without knowing whether problem `2` is occuring. Do you have any concerns on this?

KonradStaniec: I think my main concern is that, from what I understand, we copy cometbft mechanism to detect inactivity. This mechanism for sure have several invariants and assumptions inside. In cometbft/cosmos-sdk world, as soon as block `n+1` is commited, there is no more votes accepted for block `n`. 

This approach makes it possible to state at block `n+1`, that if validator have missed enough votes in window [n-50, n] , this means those votes will always be missed and this range is kind of proof of inactivity.

I am not sure we can do the same relying just on the Babylon chain tip mostly because Babylon chain tip is totally independent from finalization status. This means:
- based on chain tip we cannot make decision to stop accepting finality votes on certain height, as we do not know whether if some height eas finalized already
- but we need to stop accepting votes at some point, as otherwise finality provider will always be able to send vote after really long time, and whole inactivity tracking is kind of weird 

Lets say:
- last finalized block is 1000
- finality provider `x` was inactive in range [200, 300] and jailed, but he later unjailed
- should he be able to provide missing votes for the range [200, 300] ? In my mind he should not be able to that, and his votes for this should be rejected.

I just do not see, how mechanism proposed in this pr protects against it, and I worry that without handling this case we will break some invariants of inactivity detection mechanism copied from comet.
Does those concerns make sense or am I missing something super obvious ? 

>This forces us to consider two scenarios like you described, happy case and finality halting, and have different algorithms to the two scenarios. We also need to consider the switch of the two cases.

In my mind, this argument for me, as I usually prefer to have different scenarios and cases handled explicitly, rather than implicitly in one generic function especially as those scenarios are super different in level of severity:
- one finality provider inactive is like pretty normal things and it happens
- finality halting means something really bad is happening.

gitferry: > I just do not see, how mechanism proposed in this pr protects against it, and I worry that without handling this case we will break some invariants of inactivity detection mechanism copied from comet.
Does those concerns make sense or am I missing something super obvious ?

In the previous pr #681, we mostly adopt the inactivity detection mechanism from Cosmos-sdk where we have a bit array to track voting history (`1` indicates missed, `0` as default). We determine the value of a bit by mapping the current block height to the index of the bit in the bit array. We will not track votes for passed heights. That is, even if we received a vote for a previous height, it will not reduce the `missed_block_count` of the window.

In this PR, we introduce `liveness_delay` to determine the value of the corresponding bit of block `n` at block `n+liveness_delay` so that we allow some time to collect finality votes. After `n+liveness_delay`, votes for blocks ealier than `n` will not be counted for liveness.

> In my mind, this argument for me, as I usually prefer to have different scenarios and cases handled explicitly, rather than implicitly in one generic function especially as those scenarios are super different in level of severity:

With the proposed mechanism, we can still propose a mechanism to detect finality halting scenario and apply different policies for pubnishment but I think this could be in user story 2

KonradStaniec: ok I think I am slowly getting it with few caveats.

First, I think this comment is a bit misleading:
```
// detect inactive finality providers if there are any
// height for examining is determined by the current height - params.LivenessDelay
// this is we allow finality signatures to be received after quorum
```
as during execution of the function below we actually do not know whether quorum have been  reached at all. In strict sense,  parameter `params.LivenessDelay`, regulates have many Babylon blocks finality providers have, to cast a vote on block before being judged as missing their voting turn on given block. Once finality provider is judged that he missed his turn, this judgement is irreversible. Is this understanding correct ? 

If yes, then maybe re-name this parameter to `params.MaxFinalityVoteDelay` or `params.MaxFinalityVoteTimeout`  and adjust this comment a bit ( to underscore that this handler/parameter have nothing to do with quorum)

Another thing, I find this a bit wierd that this mechanism allows to sent votes for the past blocks even it those are finalized and finality provider was judged as inactive. I cannot find any concrete attack on that, but I do not know any blockchain system that does that. In Tendermint, after block is commited you have only `timeout_commit` time to supply votes which go above required quorum. 

Do you we should have something like that ? Like check in voting handler:
```
if blockIsFinalized(vote.height) and (current_tip - MaxFinalityVoteDelay) > vote.height:
  reject
```
This is inline with treating source blockchain block passage as clock.

gitferry: > as during execution of the function below we actually do not know whether quorum have been reached at all. In strict sense, parameter params.LivenessDelay, regulates have many Babylon blocks finality providers have, to cast a vote on block before being judged as missing their voting turn on given block. Once finality provider is judged that he missed his turn, this judgement is irreversible. Is this understanding correct ?

Exactly! Sorry for the confusion. Fixed in [e8139aa](https://github.com/babylonchain/babylon/pull/689/commits/e8139aaf7a3c07a70fb8a8413edf13745f8fd2c0).

> Another thing, I find this a bit wierd that this mechanism allows to sent votes for the past blocks even it those are finalized and finality provider was judged as inactive. I cannot find any concrete attack on that, but I do not know any blockchain system that does that. In Tendermint, after block is commited you have only timeout_commit time to supply votes which go above required quorum.

Good point. I think using babylon block as clock for finality makes sense. This makes the actual acceptance of votes inline with recording voting history. I don't see any concerns on liveness or impact on reward distribution from my side. Do you have any concerns? cc @SebastianElvis 

Nevertheless, this seems out of scope of this PR as we need to have more concrete decision of rejecting finality votes

KonradStaniec (approved):
>Exactly! Sorry for the confusion. Fixed in [e8139aa](https://github.com/babylonchain/babylon/pull/689/commits/e8139aaf7a3c07a70fb8a8413edf13745f8fd2c0).

no worries, after this discussion I think I understand better how the inactivity mechnism should work 👍

SebastianElvis:
> I don't see any concerns on liveness or impact on reward distribution from my side. Do you have any concerns?

Per yesterday's discussion. maybe we can worry about this later?

gitferry:
> Per yesterday's discussion. maybe we can worry about this later?

Yep, will probably compose a RFC for this. Merging this PR now

PR #695, opened 2024-07-09
Add `08-wasm` clients module

aeryz:
Add `08-wasm` clients to enable IBC wasm light clients: https://ibc.cosmos.network/v8/ibc/light-clients/wasm/integration/

Note that a separate wasmvm is being used for `ibc-wasm` to fully isolate `x/wasm` and `ibc-wasm` modules. Sharing VM's were [caused some issues](https://medium.com/@reecepbcups/juno-v19-post-mortum-969f251bffb7) for other people previously.


app/app.go
```diff
@@ -1296,3 +1328,16 @@ func initParamsKeeper(appCodec codec.BinaryCodec, legacyAmino *codec.LegacyAmino
 
 	return paramsKeeper
 }
+
+// Capabilities of the IBC wasm contracts
+func IBCWasmCapabilities() []string {
```
SebastianElvis, line 1333: There is a variable `wasmCapabilities` at L176 in this file. Now we could remove that

SebastianElvis (commented):
It's weird that e2e test in CI failed. Will take a look

app/app.go
aeryz: Done

SebastianElvis (approved):
Thank you!

PR #696, opened 2024-07-09
chore: Rename magic bytes to tag

gitferry:
This PR renamed `magicBytes` to `tag` to be consistent with the `tag` field in the global parameters.


RafilxTenfen (approved):
LGTM :tada:

PR #705, opened 2024-07-10
feat(gov): software upgrade governance proposal support

### LINKED ISSUE #698: feat(gov): software upgrade governance proposal support
## Overview 

Software upgrade is the process of upgrading the node software version in the middle of the execution of the blockchain. In Cosmos SDK chains, this is achieved via a [software upgrade governance proposal](https://docs.cosmos.network/main/build/modules/upgrade): one submits a proposal suggesting that the chain upgrades to a given version at a given height. If the proposal has passed, then upon this given height, the chain will halt. Only after the supermajority of validators have restarted with the given version, the chain will resume execution.

The upgrade can be breaking in the sense that the new DB schema is not compatible with the old one, e.g., a module is added/removed, DB schema has changed, etc.. In this case, the developer needs to write a migration function defining how the state under the old schema can be applied to the new schema. The migration function is versatile in the sense that one can manipulate state at wish. 

As an example, Osmosis [defines](https://github.com/osmosis-labs/osmosis/blob/3eccca25dd40ec45c0a295f079fb21da66eeeb6a/app/upgrades) such callback function for all software upgrades, and [this one](https://github.com/osmosis-labs/osmosis/blob/bdcd251208bccff2f539b7fbff3a9b7af9aa4fb4/app/upgrades/v4/upgrades.go#L12-L32) tops up OSMO tokens to a number of accounts given [here](https://github.com/osmosis-labs/osmosis/blob/8dcbaf535e9b1f9ad73878ffd1464b246edba0d1/app/upgrades/v4/prop12_data.go). Agoric has something similar [here](https://github.com/Agoric/agoric-sdk/blob/000693442f821c1fcea007a2df740733b1f75ebe/golang/cosmos/app/app.go#L865-L870.). 

## Tasks

This issue defines the task on adding software upgrade support to Babylon. We take Osmosis as the reference implementation. This includes a set of steps:

- [x] #699
- [x] #700
- [x] #701
- [x] #707

SebastianElvis:
Resolves #698 

This PR implements the feature of supporting software upgrade gov prop.

### DIFF
--- a/app/upgrades/README.md
+++ b/app/upgrades/README.md
@@ -1,0 +1,43 @@
+# Babylon Upgrades
+
+This folder contains sub-folders for every babylon upgrade. (Both state
+migrations, and hard forks) It also defines upgrade & hard fork structs,
+that each upgrade implements. These then get included in the application
+app.go to run the upgrade.
+
+The code is adapted from [Osmosis](https://github.com/osmosis-labs/osmosis/blob/68d546d94acbf1aa99d6b514cb66b2b40afff4a4/app/upgrades).
+
+## Version History
+
+TODO
+
+## Upgrade types
+
+There are two upgrade types exposed, `Upgrade` and `Fork`. An `Upgrade`
+defines an upgrade that is to be acted upon by state migrations from the
+SDK `x/upgrade` module. A `Fork` defines a hard fork that changes some
+logic at a block height. If the goal is to have a new binary be
+compatible with the old binary prior to the upgrade height, as is the
+case for all babylon `Fork`s, then all logic changes must be
+height-gated or in the `BeginForkLogic` code.
+
+```go
+type Upgrade struct {
+ // Upgrade version name, for the upgrade handler, e.g. `v7`
+ UpgradeName string
+ // Function that creates an upgrade handler
+ CreateUpgradeHandler func(mm *module.Manager, configurator module.Configurator, keepers *keepers.AppKeepers) upgradetypes.UpgradeHandler
+ // Store upgrades, should be used for any new modules introduced, new modules deleted, or store names renamed.
+ StoreUpgrades store.StoreUpgrades
+}
+
+type Fork struct {
+ // Upgrade version name, for the upgrade handler, e.g. `v7`
+ UpgradeName string
+ // height the upgrade occurs at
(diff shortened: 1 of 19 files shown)


app/upgrades/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+The code is adapted from [Osmosis](https://github.com/osmosis-labs/osmosis/blob/68d546d94acbf1aa99d6b514cb66b2b40afff4a4/app/upgrades).
+
+## Version History
+
+TODO
+
+## Upgrade types
+
+There are two upgrade types exposed, `Upgrade` and `Fork`. An `Upgrade`
```
RafilxTenfen, line 16: :100:

RafilxTenfen:
CLI tested at https://github.com/babylonchain/babylon-deployment/pull/102 
Great work :tada:

RafilxTenfen (approved):

app/upgrades/vanilla/upgrades.go
```diff
@@ -0,0 +1,61 @@
...
+	bbn "github.com/babylonchain/babylon/types"
+	btcstakingkeeper "github.com/babylonchain/babylon/x/btcstaking/keeper"
+	bstypes "github.com/babylonchain/babylon/x/btcstaking/types"
+	"github.com/btcsuite/btcd/btcec/v2"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/types/module"
+	authkeeper "github.com/cosmos/cosmos-sdk/x/auth/keeper"
+)
+
+var Upgrade = upgrades.Upgrade{
```
KonradStaniec, line 19: shoudn't this upgrade go to test code or test file ? 

If it is needed for some kind of testing, maybe we can add some comments that this is only exemple upgrade ?

gitferry (approved):

app/upgrades/vanilla/upgrades.go
SebastianElvis: If we put it under test files then we won't be able to import it in `app.go` and thus cannot test the software upgrade locally or e2e.

But yeah this shouldn't be used in production in any way. Added some docstrings and a README clarifying this.

PR #707, opened 2024-07-12
fix: execute `PreBlocker` for upgrade module

SebastianElvis:
This PR fixes a bug to execute `PreBlocker` for the upgrade module, such that the chain will handle software upgrade logic properly (i.e., at the upgrade height, the old version panicks, while the new version handles the upgrade logic and makes progress). This PR also refactors the vanilla upgrade for testing purposes.

As Babylon has a PreBlocker in the proposal handler (for BTC timestamping), this PR has taken care of this by appending checkpointing module's `PreBlocker` after the upgrade modules'.

Test plan: https://babylon-chain.atlassian.net/wiki/spaces/BABYLON/pages/340721842/PoC+for+software+upgrade+with+arbitrary+state+update. Testing this e2e requires significant amount of work that is out of this PR's scope, so suggest we do this in subsequent PRs.

### DIFF
--- a/x/checkpointing/proposal.go
+++ b/x/checkpointing/proposal.go
@@ -45,7 +45,10 @@ func NewProposalHandler(
 func (h *ProposalHandler) SetHandlers(bApp *baseapp.BaseApp) {
 	bApp.SetPrepareProposal(h.PrepareProposal())
 	bApp.SetProcessProposal(h.ProcessProposal())
-	bApp.SetPreBlocker(h.PreBlocker())
+	// extend the existing PreBlocker with the checkpoint extraction logic
+	originalPreBlocker := bApp.PreBlocker()
+	preBlocker := h.AdditionalPreBlocker(originalPreBlocker)
+	bApp.SetPreBlocker(preBlocker)
 }
 
 // PrepareProposal examines the vote extensions from the previous block, accumulates
@@ -329,14 +332,19 @@ func (h *ProposalHandler) ProcessProposal() sdk.ProcessProposalHandler {
 	}
 }
 
-// PreBlocker extracts the checkpoint from the injected tx and stores it in
-// the application
+// AdditionalPreBlocker adds checkpointing logic to the existing PreBlocker.
+// It extracts the checkpoint from the injected tx and stores it in the application
 // no more validation is needed as it is already done in ProcessProposal
-func (h *ProposalHandler) PreBlocker() sdk.PreBlocker {
+func (h *ProposalHandler) AdditionalPreBlocker(originalPreBlocker sdk.PreBlocker) sdk.PreBlocker {
 	return func(ctx sdk.Context, req *abci.RequestFinalizeBlock) (*sdk.ResponsePreBlock, error) {
-		k := h.ckptKeeper
-		res := &sdk.ResponsePreBlock{}
 
+		// execute original PreBlocker logic
+		res, err := originalPreBlocker(ctx, req)
+		if err != nil {
+			return res, err
+		}
+
+		k := h.ckptKeeper
 		epoch := k.GetEpoch(ctx)
 		// BLS signatures are sent in the last block of the previous epoch,
 		// so they should be aggregated in the first block of the new epoch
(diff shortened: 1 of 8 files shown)


x/checkpointing/proposal.go
```diff
@@ -45,7 +45,10 @@ func NewProposalHandler(
 func (h *ProposalHandler) SetHandlers(bApp *baseapp.BaseApp) {
 	bApp.SetPrepareProposal(h.PrepareProposal())
 	bApp.SetProcessProposal(h.ProcessProposal())
-	bApp.SetPreBlocker(h.PreBlocker())
+	// extend the existing PreBlocker with the checkpoint extraction logic
+	originalPreBlocker := bApp.PreBlocker()
+	preBlocker := h.AdditionalPreBlocker(originalPreBlocker)
```
KonradStaniec, line 50: It seems this is necessary because we use pre blocker a bit incorrectly. 

From what I see pre blocker should be implemented at module level. Example: https://github.com/cosmos/cosmos-sdk/blob/main/x/upgrade/module.go#L149

So then it can be run by module manager preblocker - https://github.com/babylonchain/babylon/blob/dev/app/app.go#L1078, which will execute all pre-blockers of every module.

Wdyt of fixing this on `dev` and re-basing this pr on top of dev afterwards ? (cc: @gitferry )

SebastianElvis: Fixing this is not possible at module level unfortunately. The`AppModule.PreBlocker` for a module is [defined](https://github.com/cosmos/cosmos-sdk/blob/62212dfc04b3790f3cc052cdc2fb757b32ca97d0/x/upgrade/abci.go#L25-L26) as 

```go
// HasPreBlocker is the extension interface that modules should implement to run
// custom logic before BeginBlock.
type HasPreBlocker interface {
	AppModule
	// PreBlock is method that will be run before BeginBlock.
	PreBlock(context.Context) (ResponsePreBlock, error)
}
```

while our proposal handler needs `sdk.PreBlocker` which is [defined](https://github.com/cosmos/cosmos-sdk/blob/62212dfc04b3790f3cc052cdc2fb757b32ca97d0/types/abci.go#L33-L39) as 

```go
// PreBlocker runs code before the `BeginBlocker` and defines a function type alias for executing logic right
// before FinalizeBlock is called (but after its context has been set up). It is
// intended to allow applications to perform computation on vote extensions and
// persist their results in state.
//
// Note: returning an error will make FinalizeBlock fail.
type PreBlocker func(Context, *abci.RequestFinalizeBlock) (*ResponsePreBlock, error)
```

So the module-level PreBlocker is a different thing fromo the app-level PreBlocker. Here we need app-level preblocker so the only way we can do is to wire the proposal handler's PreBlocker and the app's PreBlocker together.

KonradStaniec: ahh true, this is super wierd though given that pre-blocker at app level have `*abci.RequestFinalizeBlock` argument

Then maybe we can add our preblocker directly in `BabylonApp` in `PreBlocker` functions after module pre blockers - https://github.com/babylonchain/babylon/blob/dev/app/app.go#L1077 ?

My main reasoning to do so is that this way it will be much more visible at app wiring level and in case of any pre blocker changes (even in other modules) it will be clear that this pre-blocker is running after module preblockers ?

When this is hidden in this ` SetHandlers` it can be easily forgotten.

Not a blocker though if you have different opinion.

KonradStaniec (approved):

x/checkpointing/proposal.go
SebastianElvis: Agreed. I moved the logic of chaining two PreBlocker functions [here](https://github.com/babylonchain/babylon/blob/6874706804110377c3d4724819e539ce3e1d5644/app/app.go#L518-L531)

```go
	app.SetPreBlocker(func(ctx sdk.Context, req *abci.RequestFinalizeBlock) (*sdk.ResponsePreBlock, error) {
		// execute the existing PreBlocker
		res, err := app.PreBlocker(ctx, req)
		if err != nil {
			return res, err
		}
		// execute checkpointing module's PreBlocker
		// NOTE: this does not change the consensus prameter in `res`
		ckptPreBlocker := proposalHandler.PreBlocker()
		if _, err := ckptPreBlocker(ctx, req); err != nil {
			return res, err
		}
		return res, nil
	})
```

so that it's made explicit in `app.go`. Did not touch `SetPreBlocker` to avoid breaking the semantics of the setter

Issue #537, opened 2024-03-07
Separate database proto types from external API proto types in Babylon node


RafilxTenfen:
Currently in Babylon node in API responses we return types directly used in database:

https://github.com/babylonchain/babylon-private/blob/dev/x/btcstaking/keeper/grpc_query.go#L71 which just return delegations in straight from db

https://github.com/babylonchain/babylon-private/blob/dev/x/btclightclient/keeper/grpc_query.go#L71 which return headers as they are stored in db

This is not ideal as:

if we will need to change something in database representation (optimisation/fix/field name change) this will suddenly break our api, which at some point will expected to be stable by external consumers

not everything in db needs to be exposed to users of api.

types for api consumption may be different i.e for example db we can use []byte but in api we perfectly would use hex encoded strings

This story is about improving this situation across every module in Babylon node. The whole migration to make our node API production ready should look roughly:

Create API types in every module and write code translating between transalting db types to api types

After we have proper responses type we need to review them and decide which fields are really necessary and remove the unnecessary ones to have nice clean minimal API. We have large set or programs (dashboard/vigiilantes/reporters) to guide us in this trimming effort.

After we have nicely trimmed api, we may consolidate some practices around it so that endpoints are consistent across different modules.

PR #714, opened 2024-07-16
chore(inactivity): Rename `inactive` flag to `sluggish`, add queries, events, and metrics

gitferry:
The `inactive` flag of a finality provider is confused with the finality providers that are not in the active set. So, this PR is to rename the flag to `sluggish`, which is a common concept in the BFT world to describe validators who should vote but the votes are delayed for some reason thus committing [sluggish faults](https://dl.acm.org/doi/10.1145/3465084.3467954).

This PR also made the following additions:
* queries for signing info
* events emitted for sluggish finaltiy providers being detected or reverted
* metrics for counting the number of sluggish finality providers
* docs about the new feature


proto/babylon/finality/v1/events.proto
```diff
@@ -11,3 +11,17 @@ message EventSlashedFinalityProvider {
...
+message EventSluggishFinalityProviderDetected {
+    // public_key is the BTC public key of the finality provider
+    string public_key = 1;
+}
+
+// EventSluggishFinalityProviderReverted is the event emitted when a sluggish finality
+// provider is no longer considered sluggish
+message EventSluggishFinalityProviderReverted {
+    // public_key is the BTC public key of the finality provider
+    string public_key = 1;
```
RafilxTenfen, line 26: can we also add the bbn address of the fp to the events

x/finality/keeper/grpc_query.go
```diff
@@ -225,3 +225,51 @@ func (k Keeper) ListEvidences(ctx context.Context, req *types.QueryListEvidences
 	}
 	return resp, nil
 }
+
+// SigningInfo returns signing-info of a specific finality provider.
+func (k Keeper) SigningInfo(ctx context.Context, req *types.QuerySigningInfoRequest) (*types.QuerySigningInfoResponse, error) {
```
RafilxTenfen, line 230: Can you add a test for this and return a different structure than the one stored in the db? 
related https://github.com/babylonchain/babylon/issues/537

It could be done in another PR also ^^

RafilxTenfen (approved):
Sluggish seems funny to me kkk but I like the change :tada: 
Some suggestions but no impediments

proto/babylon/finality/v1/events.proto
gitferry: I woudn't mind but what would be the use case?

x/finality/keeper/grpc_query.go
gitferry: Nice find. Should've added tests for queries there. Done in [8be91d4](https://github.com/babylonchain/babylon/pull/714/commits/8be91d495d1476c48839851be78475432219cffe)

KonradStaniec (approved):
great work! `Sluggish` sounds a bit weird, but I agree `inactive` may be super confusing so 👍

gitferry:
> great work! Sluggish sounds a bit weird, but I agree inactive may be super confusing so 👍

Yeah, thb it's werid to me, too, but it's just temporal and will be changed to `jailed` eventually

Issue #694, opened 2024-07-08
Improve handling of parsing OP_RETURN data


KonradStaniec:
Currently op_return data for checkpoints is parsed here:
https://github.com/babylonchain/babylon/blob/dev/x/btccheckpoint/types/btcutils.go#L207

This function should clearly document that it works only for standard OP_RETRUN transactions, and return error in non standard transactions.

Standard op_return transactions:
- must have only one op_return output
- must push only up to 80 bytes of datas

PR #716, opened 2024-07-16
Improve extracting data from op_return + add tests

KonradStaniec:
Fixes: https://github.com/babylonchain/babylon/issues/694

### DIFF
--- a/x/btccheckpoint/types/btcutils.go
+++ b/x/btccheckpoint/types/btcutils.go
@@ -194,7 +194,7 @@ func verify(tx *btcutil.Tx, merkleRoot *chainhash.Hash, intermediateNodes []byte
 
 // ExtractStandardOpReturnData extract OP_RETURN data from transaction OP_RETURN
 // output.
-// If OP_RETRUN output is not standard it will be ignored. If there is more than
+// If OP_RETURN output is not standard it will be ignored. If there is more than
 // one output with OP_RETURN, error will be returned.
 func ExtractStandardOpReturnData(tx *btcutil.Tx) ([]byte, error) {
 	msgTx := tx.MsgTx()
@@ -206,15 +206,18 @@ func ExtractStandardOpReturnData(tx *btcutil.Tx) ([]byte, error) {
 		script := output.PkScript
 
 		if !txscript.IsNullData(script) {
-			// not an standard op_return, we do not care about this output
+			// not a standard op_return, we do not care about this output
 			continue
 		}
-
 		// At this point we know:
 		// - script is not empty
 		// - script is valid looking op_return
 		// - with at most 80bytes of data
 		opReturnCounter++
+
+		if opReturnCounter > 1 {
+			return nil, fmt.Errorf("transaction has more than one OP_RETURN output")
+		}
 
 		if len(script) == 1 {
 			// just op_return op code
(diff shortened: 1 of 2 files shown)


x/btccheckpoint/types/btcutils.go
```diff
@@ -204,32 +192,50 @@ func verify(tx *btcutil.Tx, merkleRoot *chainhash.Hash, intermediateNodes []byte
...
+		} else {
+			// this should be one of OP_DATAXX opcodes we drop first 2 bytes
+			opReturnData = append(opReturnData, script[2:]...)
 		}
 	}
 
-	return opReturnData
+	if opReturnCounter > 1 {
+		return nil, fmt.Errorf("transaction has more than one OP_RETURN output")
+	}
```
SebastianElvis, line 236: Can this be replaced by a nil check on `opReturnData` inside the loop? say if `txscript.IsNullData(script)` and `len(opReturnData) == 0` then return error

SebastianElvis (approved):

x/btccheckpoint/types/btcutils.go
gitferry: Are you trying to say add the following change to Line 208?

```
		if !txscript.IsNullData(script) {
			// not an standard op_return, we do not care about this output
			continue
		} else if len(opReturnData) != 0 {
                         // already found a valid opReturn data
                        return fmt.Errorf("transaction has more than one OP_RETURN output")
                }
```

Yeah, I think with the above change, we can get rid of the counter. No strong opinion though

x/btccheckpoint/types/btcutils.go
```diff
@@ -204,32 +192,50 @@ func verify(tx *btcutil.Tx, merkleRoot *chainhash.Hash, intermediateNodes []byte
...
-		// - second byte should indicate how many bytes there are in opreturn script
-		if pkScriptLen > 1 &&
-			pkScriptLen <= maxOpReturnPkScriptSize &&
-			pkScript[0] == txscript.OP_RETURN {
-
-			// if this is OP_PUSHDATA1, we need to drop first 3 bytes as those are related
+		script := output.PkScript
+
+		if !txscript.IsNullData(script) {
+			// not an standard op_return, we do not care about this output
```
gitferry, line 209:
```suggestion
			// not a standard op_return, we do not care about this output
```

x/btccheckpoint/types/btcutils.go
```diff
@@ -204,32 +192,50 @@ func verify(tx *btcutil.Tx, merkleRoot *chainhash.Hash, intermediateNodes []byte
 	return bytes.Equal(current[:], root)
 }
 
-func ExtractOpReturnData(tx *btcutil.Tx) []byte {
+// ExtractStandardOpReturnData extract OP_RETURN data from transaction OP_RETURN
+// output.
+// If OP_RETRUN output is not standard it will be ignored. If there is more than
```
gitferry, line 197:
```suggestion
// If OP_RETURN output is not standard it will be ignored. If there is more than
```

gitferry (approved):
Great fix!

x/btccheckpoint/types/btcutils.go
KonradStaniec: hmm we cannot use directly `len(opReturnData) == 0` as op_return output with empty data ie. only `OP_RETURN` opcode is an valid and standard thing. 

Although I think we can move the counter inside the loop to fail faster.

PR #723, opened 2024-07-19
epoching: test updating epoch interval

SebastianElvis:
This PR adds a test case to test updating the epoch interval in the middle of execution. In particular, it lets the chain run for a random number of blocks, update the epoch interval, ensure the current epoch metadata is not affected, and ensure the next epoch metadata is updated.

NOTE: Will try to do an e2e test for this as well.


KonradStaniec:
>NOTE: Will try to do an e2e test for this as well.

Do we need e2e test for this ? If the semantics of update is that current epoch is not affected, what more this test would give us. My main concern is that it would mostly test governance logic module logic, not epoching module logic and would add additional time to tests.

On the other hand, It would be good to see unit/integration tests which shows what is happening when parameter update execution falls on epoch boundaries i.e what happens if epoch is changed in last block of the epoch, what happens when paramter update is executed in first block of the next epoch etc.

Of course I do not have full context to this task, so this suggestions maybe not relevant  :)

PR #726, opened 2024-07-23
epoching: query CLIs for epoching module

SebastianElvis:
This PR provides query CLIs for the epoching module, including

- querying metadata of an epoch
- range query of epoch metadata
- querying queued msgs of an epoch
- querying validator set of an epoch

Test plan: Tested with a local Babylon node and the commands work.


KonradStaniec (approved):

x/epoching/client/cli/query.go
```diff
@@ -24,7 +22,215 @@ func GetQueryCmd(queryRoute string) *cobra.Command {
...
+		},
+	}
+
+	flags.AddQueryFlagsToCmd(cmd)
+	flags.AddPaginationFlagsToCmd(cmd, "epoch-validators")
+
+	return cmd
+}
+
+func getEpoch(queryClient types.QueryClient, args []string) (uint64, error) {
```
gitferry, line 211:
```suggestion
func getCurrentEpoch(queryClient types.QueryClient, args []string) (uint64, error) {
```

gitferry (approved):
LGTM!

x/epoching/client/cli/query.go
SebastianElvis: This helper function could also return the epoch number specified in `args` so let's stick to `getEpoch`

